PR #26, opened 2016-11-11
Feature/article edition bug

Crispioso:
### What

Two bugs existed existed:
1. Articles were being forced to have an edition in the create screen
2. If a space was added in the creator and then removed in the editor the 'articles' path would be removed, saving the page in an invalid location.

### How to review

Attempt to create a bulletin, article and article download with and without titles to test they give the expected result. Delete the editions in those and check the page type isn't removed from the path.

### Who can review

@robchamberspfc if you could check this one please in terms of functionality? I'll do the merge once you've given it the thumbs up


robchamberspfc:
@Crispioso Looks good locally, merge away

PR #57, opened 2017-05-04
Replace Java app with Go app

Crispioso:
### What

Replacing the original Java app with a Go application.

### How to review

Basic functionality such as starting and logging into Florence work fine. More comprehensive testing will be done on develop environment.

### Who can review

Anyone but me.


README.md
```diff
@@ -1,10 +1,53 @@
...
+| Environment variable | Default               | Description                                                                                                                              |
+|----------------------|-----------------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| BIND_ADDR            | :8080                 | Host and port to bind to                                                                                                                 |
+| BABBAGE_URL          | http://localhost:8080 | URL that Babbage can be accessed on                                                                                                      |
+| ZEBEDEE_URL          | http://localhost:8081 | The URL that Zebedee can be accessed on                                                                                                  |
+| ENABLE_NEW_APP       | false                 | When true the refactored Florence JS application will be available on `/florence`, otherwise it'll host the current Florence application |
+
+For example:
+```
+make debug BIND_ADDR=:20000
```
ian-kent: should this be `BIND_ADDR=:20000 make debug`? didn't think this way around worked

main.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	if v := os.Getenv("ENABLE_NEW_APP"); len(v) > 0 {
+		enableNewApp, _ = strconv.ParseBool(v)
+	}
+
+	log.Namespace = "florence"
+
+	/*
+		NOTE:
+		If there's any issues with this Florence server proxying redirects from either Babbage or Zebedee then the code in the previous Java Florence server might give some clues for a solution: https://github.com/ONSdigital/florence/blob/b13df0708b30493b98e9ce239103c59d7f409f98/src/main/java/com/github/onsdigital/florence/filter/Proxy.java#L125-L135
```
ian-kent: being pedantic, line could do with splitting so its readable without word wrap 😄

main.go
```diff
@@ -0,0 +1,146 @@
...
+	path := req.URL.Query().Get(":uri")
+
+	b, err := getAsset("../dist/" + path)
+	if err != nil {
+		log.Error(err, nil)
+		w.WriteHeader(404)
+		return
+	}
+
+	w.Header().Set(`Content-Type`, mime.TypeByExtension(filepath.Ext(path)))
```
ian-kent: worth looking at `http.DetectContentType`, think we've used that in other places

main_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/ONSdigital/florence/assets"
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestMain(t *testing.T) {
+	Convey("Returns 200 when asset is requested", t, func() {
+		recorder := httptest.NewRecorder()
+		rdr := bytes.NewReader([]byte(``))
+		request, err := http.NewRequest("GET", "/florence/dist/js/florence.bundle.js", rdr)
```
ian-kent: can pass in `nil` instead of `rdr` if the body has 0 bytes

ian-kent:
Only reviewed the go app, ignoring the refactored JS code since it won't be used anyway. Probably worth @LloydGriffiths having a look at the Jenkinsfile

README.md
Crispioso: That does work (just tested it) but I'm happy to change it if it's not what others would expect for environment variables.

ian-kent: naahhh

main.go
Crispioso: Good shout!

main_test.go
Crispioso: 👍

main.go
Crispioso: That doesn't detect `.css` properly so will stick with current solution unless there's a benefit to `http.DetectContentType`.

Crispioso:
@ian-kent Lloyd has said it all looks good to go. So I think we're ready to approve this.

PR #63, opened 2017-05-16
increase default server timeouts

ian-kent:
### What

Replaced default timeouts from go-ns server

* IdleTimeout = 120s (time to hold idle connections without requests)
* WriteTimeout = 120s (time to complete write)
* ReadTimeout = 30s (time to read end of request)

### How to review

* Hack zebedee to introduce a 60s delay on some request, e.g. collection list
* Make sure it still works in Florence

### Who can review

Anyone except @ian-kent


LloydGriffiths:
LGTM, probably worth @Crispioso or @jondewijones taking a look.

PR #72, opened 2017-07-13
Improve client-side logging

Crispioso:
### What

Improved some of the logging in Florence to help us when a content loss issue happens:
- Add logging on 'GET content' to help get coverage on what information is coming back from Zebedee after a 'POST content'.
- Log timestamps on post content, get content, disable Florence UI and enable Florence UI.
- Remove ping request results from the console and store in a global variable.

### How to review

1) Switch to the branch `feature/logging`
2) Open a bulletin in the editor and check the GET and POST request for the page content are logged.
3) Pings should no longer show in the console but are available with the `Florence.ping` global object.

### Who can review

Anyone but me.


dist/legacy-assets/js/main.js
```diff
@@ -295,6 +308,9 @@ function getPageData(collectionId, path, success, error) {
     dataType: 'json',
     type: 'GET',
     success: function (response) {
+      var date = new Date();
+      date = date.getHours() + ":" + date.getMinutes() + "." + date.getMilliseconds();
```
ian-kent: this line is repeated quite a lot - could be worth creating a log function which does it for you?

dist/legacy-assets/js/main.js
```diff
@@ -11539,34 +11560,27 @@ function setShortcuts(field, callback) {
...
 
-                if (averagePingTime < 100) {
-                    console.log("ping time: pretty good! " + time + " average: " + averagePingTime + " " + pingTimes);
-                } else if (averagePingTime < 300) {
-                    console.log("ping time: not so good! " + time + " average: " + averagePingTime + " " + pingTimes);
-                } else {
-                    console.log("ping time: really bad! " + time);
-                }
+                var date = new Date();
+                Florence.ping.add(time, date)
```
ian-kent: should `add` need a time?

Crispioso: Good point, resolved 👍

dist/legacy-assets/js/main.js
Crispioso: Perhaps worth doing if we use it more often? We haven't got any global logging function yet but could be worth adding it into something like that if we find ourselves putting more time into it.

PR #76, opened 2017-08-09
Update README

Crispioso:
### What

Updated the README with some CMD information and extra details about what other applications are required to get Florence running end-to-end.

### How to review

Quick check of the README should suffice.

### Who can review

Anyone but me.


Crispioso:
@mattrout92 Good point, I've added those into the README too.

PR #83, opened 2017-08-29
Add healthcheck to florence

mattrout92:
### What

Add healthcheck to florence

### Who can review

Anyone


ian-kent:
Couple of thoughts:

- We should think about graceful degradation, e.g. failing a healthcheck will cause florence to become 100% unavailable, but actually if zebedee works ok but the recipe API doesn't, we could offer a degraded service instead of taking all of Florence offline

- We should make the healthchecks run concurrently (in go-ns) - the current configuration, with a 5 second timeout per healthcheck and a 60 second iteration, means with 12+ healthchecks, the loop could start stacking up (could be avoided by waiting for 60 seconds _between_ healthcheck iterations rather than starting the timer before running the healthcheck)

- We should be using an RWMutex in go-ns healthcheck - at the moment, any request to 'Do' will block until any in-progress healthchecks complete (also, we should create a temporary healthState and only lock the mutex around the assignment back to the package level copy)

- Potentially worth considering how we _stop_ healthchecks running - typically that would be done at app shutdown, but the current loop can never exit (could be linked to an exit channel of some kind which can be closed on app shutdown/os signal)

PR #112, opened 2017-10-12
Feature/select a dataset

mattrout92:
### What

Update select a dataset page to match UX designs

### How to test

Make sure you have an instance (or instances) that are completed and match to a dataset id in mongodb. Visit http://localhost:8081/florence/datasets and verify functionality works


src/app/views/datasets/DatasetsController.jsx
```diff
@@ -113,58 +108,74 @@ class DatasetsController extends Component {
...
-                          isUpdating={this.state.isFetchingDatasets}
-                          handleItemClick={this.goToDatasetDetails}
-                          items={this.state.allDatasets}
-                      />
+                <div className="grid__col-7">
+                    <ul className="list list--neutral">
+                        <li className="list__item grid grid--justify-space-between">
+                            <h1>Select a dataset</h1>
+                            <div className="margin-top--3">
+                                <a className="btn btn--primary" href={`${location.pathname}/uploads`}>Upload a dataset</a>
```
Crispioso: This should use the <Link> component rather than an <a> element (eg https://github.com/ONSdigital/florence/blob/099a9bd7f9e5e6c377484ae822a8c0049d549085/src/app/views/datasets/DatasetsController.jsx#L153). Using the <a> causing the page to do a full page reload, rather than rendering the next route with React without a hard load.

src/app/views/datasets/selectable-table/SelectableTableController.jsx
```diff
@@ -0,0 +1,119 @@
...
+            this.setState({table: array});
+        }
+    }
+
+    render() {
+        return (
+            <div className="selectable-table">
+                    <div className="selectable-table__heading">
+                        <ul className="selectable-table__list">
+                            <li className="selectable-table__list-item grid__col-6">Title</li>
```
Crispioso: Because these share the class `selectable-table__list-item` they have a hover state which shows a pointer, which suggests it's clickable. Perhaps use a different classes for the heading list and items?

src/app/views/datasets/DatasetsController.jsx
```diff
@@ -113,58 +108,74 @@ class DatasetsController extends Component {
...
-            const instanceDataset = datasets.find(dataset => {
-                return dataset.id === instance.dataset_id
-            });
-            return {
-                id: instance.id,
-                name: instanceDataset.title || `No name available (${instance.id})`
-            }
-        });
-    }
+        instances.map(instance => {
```
Crispioso: I currently get an empty list because I have 1 dataset on my local but no instances, because we loop round instances instead of datasets. The solution to this might be reworking this loop so it goes through datasets instead, but this probably needs a bit of extra thought.

PR #141, opened 2017-11-27
Unstub dataset metadata

Crispioso:
### What

Unstub the dataset metadata screen so that it is using the API and storing data properly.

### How to review

Go to a dataset metadata screen, on most people's local this will be:
`/florence/datasets/931a8a2a-0dc8-42b6-a884-7b6054ed3b68/metadata`

You should be able to edit and save all of the information on this screen.

### Who can review

Anyone but me.


src/app/utilities/http.js
```diff
@@ -15,7 +15,7 @@ export default class http {
     * @returns {Promise} which returns the response body in JSON format
     */
     static delete(uri, willRetry, callerHandles401) {
-        return request("DELETE", uri, willRetry, null, undefined, callerHandles401);
+        return request("DELETE", uri, willRetry, undefined, undefined, callerHandles401);
```
mattrout92: changing this to undefined appears to break florence login - get error message

PR #171, opened 2018-02-08
Feature/metadata additional fields

boxadesign:
### What

Added additional fields and ux changes to dataset and version metadata pages. 

### How to review

All fields should now work correctly with the dataset api. Check that all fields can be created, saved, changed etc.

Note - currently unable to delete any data that's stored as an array i.e. alerts, changes. API change is required.

### Who can review

Anyone.


Crispioso:
Looks good, the metadata screens have come a long way in this PR! A few things to note, not for you alone, but for all of us to come back to at some point:

- If you edit a dataset, add it to a collection, then preview, the `back` link takes you to the version metadata screen. Should it take you back to the dataset metadata screen?
- If you edit the dataset title the 'Dataset title` heading also changes. Do we want to keep this as the original and only update it once it's been saved?
- I'm getting a warning somewhere (I think the dataset metadata screen?) about an uncontrolled input.
- I think we need to make lots of choices about what inputs are disabled. For example, should I even get the option to change an edition input back to empty after it's been selected? Also, should we scroll users to the top of the page where there are all of the validation errors listed, because it could be off screen if not. Etc etc etc.
- I got a 404 when PUTting versino metadata changes, I think due to my own local issues, but there were no errors messages and I went to the datasets screen without knowing that anything hadn't worked. We probably need to show some notification to show something when wrong.


I think it's worth merging this PR in and just making tickets / having a chat with UX about what we want to do with the above comments It feels like we want to know whether these screens actually work for the users before we start polishing them / bug fixing.

PR #189, opened 2018-03-01
Feature/update dimension titles

boxadesign:
### What

Changed dimension titles to be editable inputs on the version metadata screen.

### How to review

Check that the dimension name is the input's value if no label exists.
Check the label is used if one exists.
Check that when the input value is changed, the label field is updated in the dimension data upon save.
Check that the new label is displayed upon reloading the page.

### Who can review

Anyone.


Crispioso:
Something to consider: we might want the dimension name inputs to behave like other inputs when the value is removed. At the moment if you highlight the input text and delete it then it automatically reverts to the original dimension name - we possibly want to allow it to behave like other inputs. I.e. the user can empty it but it causes a validation error and we then show an inline error in the invalid input/s.

PR #197, opened 2018-03-08
Feature/nav bar working on section

jondewijones:
### What

- Added global property with workingOn property
- Set active global collection and cookie when collection is selected
- Added "working on" section to nav bar (appears only when editing edition/versions)
- Refactored navbar component 

### How to review

Code review and test

### Who can review

Anyone


Crispioso:
This all works great 👍

PR #200, opened 2018-03-19
Feature/add to collection on save

Crispioso:
💚 [dataset review/approve workflow](https://github.com/ONSdigital/florence/pull/198) is now merged, so this is good to GO! 💚  

### What

On the first save of dataset or instance that item should be added to 'in progress' in the current colleciton.

### How to review

1) Switch Zebedee to `feature/datasets-review-workflow`.
2) Select 'Add dataset' in a collection.
3) Go to a dataset metadata screen and save.
4) Go back to active collection and it should be in progress now.
5) Do the same with an instance (this shouldn't be possible with a version because to become a version it has be to added to a colleciton).

### Who can review

@jondewijones


jondewijones:
Approved but when datasets are first added to a collection no title is returned from the collectionDetail end point so shows as `undefined`

Crispioso:
I'm not able to reproduce the `undefined` dataset title bug at the moment. We'll get this merged in and then check/fix that bug.

PR #240, opened 2018-05-02
Repeat check for filter downloads in florence preview to remove downl…

mattrout92:
### What

Update JS to check for filter downloads in florence preview and remove download service domain if they exist

### How to review

Generate filter downloads in florence preview and verify you can download them!

### Who can review

Anyone

### DIFF
--- a/src/app/views/datasets/preview/DatasetPreview.jsx
+++ b/src/app/views/datasets/preview/DatasetPreview.jsx
@@ -34,21 +34,33 @@ class DatasetPreview extends Component {
         // The downloads need to be sent through the florence domain so the florence
         // token can be sent to the download service to authenticate the private
         // download links
+        function updateDownloadLinks() {
+            var iframe = document.getElementById('iframe');
+            if (iframe != null) {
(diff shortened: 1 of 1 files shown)


src/app/views/datasets/preview/DatasetPreview.jsx
```diff
@@ -34,21 +34,33 @@ class DatasetPreview extends Component {
...
-            csvDownload.setAttribute("href", csvHref);
+                var excelDownload = innerDoc.getElementById('excel-download');
+                if (excelDownload != null) {
+                    var excelHref = excelDownload.getAttribute("href").replace(/^.*\/\/[^/]+/, '');
+                    excelDownload.setAttribute("href", excelHref);
+                }
+            }
         }
+        
+        setInterval(updateDownloadLinks, 1000);
```
CarlHembrough: I don't understand the reason for the interval - was the being executed too soon? Might be worth asking one of the front end devs if there is an explicit even we can hook onto instead of having an interval

PR #242, opened 2018-05-10
Allow cookies to be accessible through subdomains

mattrout92:
### What

- Allow cookies to be accessible through subdomains (e.g download service would be `download.publishing.cmd.onsdigital.co.uk`)
- Remove javascript which forced download service requests through florence
- Remove download service url from config

### How to review

Verify that the logic is correct

### Who can review

Team A/ Team B


src/app/views/datasets/preview/DatasetPreview.jsx
```diff
@@ -30,26 +30,7 @@ class DatasetPreview extends Component {
...
-            csvDownload.setAttribute("href", csvHref);
-        }
-
-        var xlsDownload = innerDoc.getElementById('xls-download');
-        if (xlsDownload != null) {
-            var xlsHref = xlsDownload.getAttribute("href").replace(/^.*\/\/[^/]+/, '');
-            xlsDownload.setAttribute("href", xlsHref);
-        }
-    }
+    handlePreviewLoad() { }
```
Crispioso: I think we should be able to remove this declaration now?

src/app/views/datasets/preview/DatasetPreview.jsx
```diff
@@ -30,26 +30,7 @@ class DatasetPreview extends Component {
         this.handlePreviewLoad = this.handlePreviewLoad.bind(this);
```
Crispioso: I think we should be able to remove this binding now it doesn't do anything?

PR #256, opened 2018-06-27
added url2 to bulletins

Richard-95:
### What

Modified interactive to handle url2 variable.
https://trello.com/c/24GgqjnO/1211-the-ability-for-data-vis-to-have-a-side-by-side-i-frame
https://github.com/ONSdigital/babbage/pull/233
https://github.com/ONSdigital/sixteens/pull/125

### How to review

create a bulletin in florence, add content "url" and "url2" and check embed code. 

### Who can review

jon - crispin


src/legacy/js/functions/_loadEmbedIframe.js
```diff
@@ -15,6 +15,8 @@ function loadEmbedIframe(onSave) {
     function saveUrl() {
         var embedUrl = $('input#embed-url').val();
         var fullWidth = $('input#full-width-checkbox').is(':checked');
+        var embedUrl2 = $('input#embed-url-2').val();
+        console.log(embedUrl2," - 1");
```
Crispioso: Oops, looks like there might be a rogue `console.log` left. Unless we think this is useful in the production environment it's probably worth deleting.

src/legacy/scss/partials/_shame.scss
```diff
@@ -43,6 +43,9 @@
         z-index: 1001;
         width: 500px;
         height: $lineheight * 17;
+        &--height-50{
```
Crispioso: Out of interest, is there a reason for this being `50` specifically? It could be handy to be a bit more descriptive with the name, for example is a larger or taller modal rather than `50`?

src/legacy/scss/partials/_shame.scss
```diff
@@ -43,6 +43,9 @@
...
+            height: 400px;
```
Crispioso: Could we possibly use the `$lineheight` variable instead hardcoding the height, to keep things as consistent as possible? For example
```
height: $lineheight * 21;
```

PR #276, opened 2018-07-18
Feature/trim whitespace on inputs

jondewijones:
### What

Added function to remove whitespace from both sides on the value

### How to review

Try adding whitespace to inputs in Florence. Whitespace should be removed on blur

### Who can review

Anyone


src/legacy/js/functions/_setupFlorence.js
```diff
@@ -371,5 +371,15 @@ function setupFlorence() {
             this.collection = collectionTemp
         }
     }
+
+    $(document).on('blur', 'input, textarea', function() {
+        if (this.type === "file") {
```
Crispioso: This if statement doesn't cover a select, checkboxes and other input types that have return a user input value so don't need to run the same functions.

Crispioso: Ignore me, I've found out this was only done on file inputs because it causes a runtime error. I'll add a comment to the code so we know in the future though.

PR #278, opened 2018-07-30
Feature/preview

jondewijones:
### What

- Build preview component
- Allow users to login and view collections list

### How to review

Setup a viewer account, that is part of a team that can access a collection that has content. Use the viewer account to login and test

### Who can review

Probably @Crispioso


Crispioso:
Leaving this here as a note of things that I need to investigate or fix:
1) Going to `/florence/preview` shows 'This page does not exist' but shows the drop-down. Should it redirect to the login screen if I'm not logged in or redirect to the collections screen if not collection ID is given?
2) Should the navbar show what collection is being previewed?

PR #288, opened 2018-09-11
Fix padding and hover/focus styling on password input

Crispioso:
### What

Fix two things:
- Long values in the password input (or with a small screen) the show/hide button no longer covers over the input text.
- Remove the focus outline (we already apply an underline on focus) on the show/hide button.
- Add cursor pointer on hover of the show/hide button.

### How to review

Going to the login screen and check that the styling bugs above are fixed.

### Who can review

Anyone but me.


sully90:
Still getting the issue where the show/hide covers the text. Haven't run Florence in a while but did a pull and checked out this branch, still see it. Is there anything else I should have done?
screen shot 2018-09-11 at 18 11 26

PR #296, opened 2018-10-02
CMD functionality (feature flagged)

Crispioso:
### What

Merge the CMD functionality into the main environments. However, all CMD functionality is feature flagged and should be hidden by default.

### How to review

1) Run `make node-modules`,
1) Check that Florence works as it currently does on develop. I.e. collections can be created, pages can be saved etc.
2) Check that no CMD functionality displays through the app, for example the 'Datasets' tab in the nav bar and the routes aren't accessible.
3) Check that setting the environment variable `ENABLE_DATASET_IMPORT=true` shows the CMD functionality (the functionality won't work without the full CMD stack but it should still display and be accessible through Florence).

### Who can review

Anyone but me.


Crispioso:
This PR  had accidentally broke the commit history with the commit `b4c3835`. A new PR has been opened with the correct commit history.:
https://github.com/ONSdigital/florence/pull/301

PR #330, opened 2019-01-29
Feature/update server logging

ian-kent:
### What

Replace [go-ns/log](https://github.com/ONSdigital/go-ns/tree/master/log) with [log.go](https://github.com/ONSdigital/log.go/tree/master/log)

### How to review

Run florence and check that logs have the correct structure

### Who can review

Anyone except @ian-kent


main.go
```diff
@@ -343,52 +331,19 @@ func websocketHandler(w http.ResponseWriter, req *http.Request) {
...
-		// if err != nil {
-		// 	log.ErrorR(req, err, nil)
-		// 	break
-		// }
-	}
-}
-
-// FIXME this is to fix go-ns reverse proxy replacing host name so that
-// babbage has correct values when using {{location.hostname}} in handlebars
-func createBabbageProxy(proxyURL *url.URL, directorFunc func(*http.Request)) http.Handler {
```
daiLlew: Is this meant to have been removed?

ian-kent: yeah, it now uses reverseProxy from go-ns, and it's the frontend-router now instead of babbage

daiLlew: 👍 just checking.

PR #338, opened 2019-02-12
logging in users screen

Richard-95:
### What

added logging to users screen

### How to review

open chrome developer tools Network tab
websocket Frames
Create/delete users
logs successfully created/deleted users

### Who can review

jon


jondewijones:
It would be good to add some more logs where we don't have any e.g. form validation errors, mapper methods, etc.

PR #341, opened 2019-02-14
validation of characters in collection name

Richard-95:
### What

Added validation of characters in collection name for special characters

### How to review

goto florence/collections
Create a collection with special characters in the name
Create collection
get error "Collection names can only contain letters and numbers. (special characters) are not allowed."

### Who can review

all


src/app/views/collections/validation/collectionValidation.js
```diff
@@ -7,17 +7,27 @@
...
     static name(name) {
+        let response = {
+            isValid: true,
+            errorMsg: ""
+        }
+
+        if (name.match(/[^a-zA-Z0-9 ]/)) {
+            response = {
+                isValid: false,
+                errorMsg: "Collections must not have (" + name.replace(/[a-zA-Z0-9 ]/g, "") + ") symbols"
```
jondewijones: It might be better to change this to a positive or guiding message. E.g how to fix the error, as in what can be used "A-Z, a-z, 0-9". It could be helpful to have what characters are causing the error, but i'd put those after how to fix it

jondewijones:
We'll need to add some unit tests for this logic

PR #345, opened 2019-03-01
 Set domain when trying to delete/expire cookie

jondewijones:
### What

Set domain when trying to delete/expire cookie. Just a test to see if this fixes cookies not being removed in environments as it works locally (before and after this change).

### Who can review

Anyone


src/app/utilities/cookies.js
```diff
@@ -54,7 +54,8 @@ export default class cookies {
             console.error(`Unable to find '${name}' cookie`);
             return false; 
         }
-        document.cookie = name + '=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;';
+        document.cookie = `${name}=;path=/;domain=${location.hostname};expires=Thu, 01 Jan 1970 00:00:01 GMT;'`
+        //document.cookie = name + '=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;';
```
janderson2: We should probably remove this commented out bit.

PR #347, opened 2019-03-15
Feature/load previous dimensions for new versions

jondewijones:
### What

- Add logic to read last published dimensions for versions that are in `edition-confirmed` state
- Updated existing logging to new logging package to enable adding new logging to new functionality

### Who can review

Anyone


src/app/views/datasets-new/edit-metadata/DatasetMetadataController.jsx
```diff
@@ -176,7 +176,14 @@ export class DatasetMetadataController extends Component {
                 isGettingVersionMetadata: false, 
                 versionIsInCollection: mappedVersion.collection, 
                 instanceID: mappedVersion.instanceID, 
-                versionIsPublished: mappedVersion.versionIsPublished});
+                versionIsPublished: mappedVersion.versionIsPublished
+            });
+
+            // if version state is edition-confirmed load in dimensions  
+            // labels and descriptions from last published version
+            if (version.state === "edition-confirmed") {
```
daiLlew: Picky comment but best practice would be to have the constant as the first arg in the comparison as it's safer if version is nil.
```
if ("edition-confirmed" === version.state) { ... } 
```

PR #353, opened 2019-04-16
Removed validation of release date

dandwal:
### What

* isReleaseDateEnabled has now been added as an option for neutral Articles
* when creating neutral articles the user of the CMS is given an option to enable or disable the release date
* When editing article release date enabled option is present 
* isReleaseDateEnabled is written to file

### How to review

1. Checkout the branch related to the following pull request in Babbage: https://github.com/ONSdigital/babbage/pull/310
2. Checkout the following branch in Florence:  feature/neutral-article-release-date-opt
3. Check that when you create and edit an article there is an option to enable/disable the release date.
4. Check that when you create and edit any page other than an article you are not presented with the option of enabling and disabling the release date.
5. When editing/creating an article ensure that the Enable Release date option follows the following rules:
* If it is not a neutral article then release date enabled box should be checked. And should not be clickable to change its state.
* If it is a neutral article then the checkbox for enabling release date should become clickable and change its state based on the true value of the enable release date boolean flag which is visible if you open the pages JSON file.

6. Check that the Babbage side works
### Who can review

Anyone except me on the Dev team


jondewijones:
All the functionality looks good. However, there seems to be a problem with how the article editor is rendered (probably a markdown issue in the handlebars)

When toggled to "neutral" article the template works fine, but when in normal article mode breaks. 
What it should look like: (when toggled to "neutral")
Screenshot 2019-04-26 at 14 25 15

What it looks like currently: (when toggled to normal article)
Screenshot 2019-04-26 at 14 25 02

PR #355, opened 2019-05-02
Feature/collection route wrapper

jondewijones:
### What

Create a collection route wrapper to handle any collection level logic in a single place that wraps all routes that can edit content within a collection. 

### How to review

This doesn't do a lot right now, but checks if collection data is already in state, and if not adds it. This can be checked by going directly to a collection content edit page such as edit a dataset e.g. `/collections/<<collectionID>>/datasets/<<datasetID>>` and check collection data is loaded into redux state.

### Who can review

Anyone


src/app/config/actions.js
```diff
@@ -127,10 +127,10 @@ export function emptyActiveCollection() {
     }
 }
 
-export function updateWorkingOn(id, name) {
+export function updateWorkingOn(id, name, url, collection, error) {
```
dandwal: 1. Might be worth considering changing these parameters to one object that has these as properties. If for example, we needed to add a new argument to this function we would be forced to supply, `id`, `name`, `url` `collection` and `error` before our new arguments, or be forced to change the order of the arguments and then every place where these are called. 

e.g. 
```
function updateWorkingOn({id, name, url, collection, error}) {...
```
Which can then be called like `updateWorkingOn({id:a, name:b, error:c})` etc. 

It would just make it easier to maintain and keeps the function open for expansion in the future.

2. I know this was the name of the function already, IMO with just the context of it is in a file named 'actions.js' and function is called `updateWorkingOn` I have no clue what the purpose of this function was without seeing where it was called from. This to me is a sign that annotations are needed, although IMO the whole file should have annotations for each function. If against annotations and opting for self annotating code, then these event names need to be prefixed with something like 'eventUpdateStatusToWorkingOn' (or do both which would be even better).

3. Similar to point 2 but with the argument names. There are no annotations and the names are not declarative enough e.g. 'url' url of what?. This just makes it harder to maintain in the future, and harder for starters to get to grips with the code.

src/app/global/collection-wrapper/CollectionRoutesWrapper.jsx
```diff
@@ -0,0 +1,134 @@
...
+        pathname: PropTypes.string.isRequired,
+    }),
+    rootPath: PropTypes.string.isRequired,
+    workingOn: PropTypes.shape(),
+    children: PropTypes.element
+};
+
+export class CollectionRoutesWrapper extends Component {
+
+    componentDidMount = () => {
```
dandwal: Shouldn't this be a verb; for the name of a function?

src/app/global/collection-wrapper/CollectionRoutesWrapper.jsx
```diff
@@ -0,0 +1,134 @@
...
+            return;
+        }
+        if (this.props.workingOn) {
+            return;
+        }
+        this.getCollectionDetails(collectionID);
+    }
+
+    getCollectionDetails = collectionID => {
+        return collections.get(collectionID).then(response => {
```
dandwal: Might be worth replacing promises (.then) with async/await

src/app/global/collection-wrapper/CollectionRoutesWrapper.jsx
```diff
@@ -0,0 +1,134 @@
...
+        }).catch(error => {
+            switch(error.status) {
+                case(401): {
+                    // do nothing - this is handled by the request function itself
+                    break;
+                }
+                case(404): {
+                    const notification = {
+                        type: 'neutral',
+                        message: "Collection couldn't be found so you've been redirected to the collections screen",
```
dandwal: Collection -> 'The collection'

src/app/global/collection-wrapper/CollectionRoutesWrapper.jsx
```diff
@@ -0,0 +1,134 @@
...
+        log.event("collection response to 'working on' state");
+        try {
+            return {
+                id: collection.id,
+                name: collection.name,
+                url: `${this.props.rootPath}/collections/${collection.id}`,
+                error: false
+            }
+        } catch (error) {
+            log.event("error mapping collection response to 'working on' state", log.error(error),
```
dandwal: Might be worth using the same casing of 'error/Error' in copying the same error message for `log.event` and `console.error`

src/app/global/collection-wrapper/CollectionRoutesWrapper.jsx
```diff
@@ -0,0 +1,134 @@
...
+            console.error("Error mapping collection response to 'working on' state", error);
+            const notification = {
+                type: 'warning',
+                message: "An unexpected error occurred whilst getting details about this collection, please refresh the page",
+                autoDismiss: 5000
+            };
+            notifications.add(notification);
+            return {
+                error: true
+            }
```
dandwal: A bunch of missing ';'

src/index.js
```diff
@@ -92,18 +93,20 @@ class Index extends Component {
...
+                                    <Route path={`${rootPath}/collections/:collectionID/datasets`} >
+                                        <IndexRoute component={ userIsAuthenticated(SelectADataset) }/>
+                                        <Route path=':datasetID'>
+                                            <IndexRoute component={ userIsAuthenticated(DatasetEditionsController) }/>
+                                            <Route path={`editions`} component={ userIsAuthenticated(CreateEditionController) }/>
+                                            <Route path='editions/:editionID'>
+                                                <Route path={`instances`} component={ userIsAuthenticated(CreateVersionController) }/>
+                                                <IndexRoute component={ userIsAuthenticated(DatasetVersionsController) }/>
+                                                <Route path={`versions/:versionID`} component={ userIsAuthenticated(DatasetMetadataController) }>
+                                                    <Route path={`edit/:metadataField/:metadataItemID`} component={ userIsAuthenticated(EditMetadatItem) }/>
```
dandwal: Typo: EditMetadatItem -> EditMetadataItem (initial type wasn't on this PR but just noticed it, also on the import itself)

There is also another in this file just noticed but is out of the scope of PR (feel free to update it though :P 'userisAdminOrEditor' -> 'userIsAdminOrEditor'

src/app/global/collection-wrapper/CollectionRoutesWrapper.jsx
jondewijones: Do you mean `componentDidMount`? `componentDidMount` is a React lifecycle method.

src/app/global/collection-wrapper/CollectionRoutesWrapper.jsx
jondewijones: Will get fixed when we add a formatter

src/app/global/collection-wrapper/CollectionRoutesWrapper.jsx
jondewijones: Fixed

src/app/global/collection-wrapper/CollectionRoutesWrapper.jsx
jondewijones: FIxed

src/app/global/collection-wrapper/CollectionRoutesWrapper.jsx
dandwal: My limited knowledge of React showing here :D

src/index.js
jondewijones: Fixed

src/app/global/collection-wrapper/CollectionRoutesWrapper.jsx
```diff
@@ -0,0 +1,134 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { replace } from 'react-router-redux';
+
+import { updateWorkingOn } from '../../config/actions'
```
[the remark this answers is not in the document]
jondewijones: Will get fixed by formatter

src/app/config/actions.js
dandwal: Discussed, no action to be taken.

PR #364, opened 2019-06-06
Fix/viewer see completed collection

dandwal:
### What

* Viewers should see completed unpublished collections
* Tests added for above 

### How to review

1. Create a collection
2. Add contents to the collection, 
3. Review all contents of the collection
4. Approve collection 
5. Note this should move to publish queue
6. Note this should no longer be visible on the select collection screen
7. Change user to a viewer (make sure this viewer is in a team that has access to the collection
8. Note the collection should be visible on the select collection screen

### Who can review

Anyone except me


src/app/views/collections/CollectionsController.test.js
```diff
@@ -311,4 +321,15 @@ describe("mapStateToProps function", () => {
     it("returns the application's root path", () => {
         expect(mappedProps.rootPath).toBe("/florence");
     });
+});
+
+describe("Viewer user types are able to see all collections, even completed", () => {
```
daiLlew: Possibly consider adding an `expect` statement on on the collection state as well - just to be sure.

Also I'm not sure where the mocked collections come from (so dismiss this if it's nonsense) but in the mocked collections list used by these tests is there a mix of collections in various states _i.e._ 3 collections with 2 in progress 1 completed?
I'm sure it's fine but I think it might be worth while adding a few different cases to verify this does work as expected.

dandwal: Thanks for the review, yep definitely worth having multiple states of collections to test this on, in this case, there were a set of mock collections available made up of 5 collections each with different states/properties. I am pushing some improvements now.

PR #370, opened 2019-07-10
Fix/mapping dataset version collection state

jondewijones:
### What

Add dataset id check to version collection state mapper method. 

### How to review

Add two datasets (e.g. `cpih01` and `mid-year-pop-ests`) and two versions to the same collection. Make sure the version numbers are the same, and editions have the same ID. You should be able to get all content into "approved" and be able to approve the collection. It should still work when version numbers and editions are different. 

### Who can review

Anyone


src/app/views/datasets-new/edit-metadata/DatasetMetadataController.jsx
```diff
@@ -365,7 +365,10 @@ export class DatasetMetadataController extends Component {
     mapVersionCollectionStateToState = versions => {
         try {
             const version = versions.find(version => {
-                return version.version === this.props.params.versionID && version.edition === this.props.params.editionID;
+                console.log(version)
+                return version.version === this.props.params.versionID &&
```
daiLlew: Can `this.props` or `this.props.params` ever be nil?

jondewijones: Yes if the user tampers with the URL, but it would have thrown an error (displayed to the user) before getting to this method.

daiLlew: 👍

PR #372, opened 2019-07-26
Add CI formatting checks

dandwal:
### What

* CI now checks that source code has been correctly formatted
* Moved formatter config to its own file prettierrc.json

### Who can review

Anyone except me


ci/scripts/unit.sh
```diff
@@ -3,5 +3,8 @@
 export GOPATH=$(pwd)/go
 
 pushd $GOPATH/src/github.com/ONSdigital/florence
-  make test
+make test
+  if  prettier --check "src/app/**/*.js" | tee /dev/stderr | grep -q "Code style issues found"; then
```
gedge: just out of interest, I looked at this, and found on the [website](https://prettier.io/docs/en/cli.html#check):

> The command will return exit code 1 in the second case, which is helpful inside the CI pipelines

so, you could just use (instead of these three lines):

```
prettier --check "src/app/**/*.js" || exit $?
```

gedge: thinking about it, you might want to put that test into the `Makefile` instead of this file:

```
test-pretty:
            prettier --check "src/app/**/*.js"
```

and consider adding that target to the `test` target in there, too:

`test: test-go test-npm test-pretty`

gedge:
I'm happy with the CI/makefile side of this - I assume that @jondewijones can check the JS malarky. 🤷‍♂

PR #376, opened 2019-09-03
Added filterable publications to published report screen

dandwal:
### What

Added filterable publications to published report screen (only path has been added) as seen in the image below:
Screenshot 2019-09-03 at 13 26 21

_Note: Comments have been placed over the PR to explain design decisions_

#### Dependencies:

- Zebedee PR: https://github.com/ONSdigital/zebedee/pull/297

### How to review

1. Checkout this PR and dependencies listed above
2. Check the source code for design issues or errors
3. Publish a dataset (or a few) along with a version of an edition (or a few)
4. Publish some BAU pages
4. Check out the reports screen (Note if testing locally you will need to be running a specific version of dp-compose found here: https://github.com/ONSdigital/dp-compose/tree/feature/postgres-version (Note that you should not be running elastic-search outside of dp-compose or the report page will likely still not work)
5. Ensure all the BAU reports are accurate as before and that the new reports are also accurate

### Who can review

Anyone except me

### DIFF
--- a/dist/legacy-assets/js/main.js
+++ b/dist/legacy-assets/js/main.js
@@ -55782,7 +55782,40 @@ function viewReleaseSelector() {
 
 function viewReportDetails(collection, isPublished, $this) {
 
-    var details, reportDetails, published, events;
+    // Enum type of filterable data
+    const filterableReportType = {
+        'dataset': 0,
+        'datasetVersion': 1
+    };
+
+    // Retrieve all information for a specific type of filterable data report
+    function getFilterableReportData(filterableDataTypeToUse, fullDatasetCollectionData) {
+        let listOfReportData = [];
+
+        // Retrieve all mandatory information fields for filterable data report
+        function getMandatoryFilterableReportInformation(datasetCollectionItem) {
+            const url = new URL(datasetCollectionItem.uri);
+            return {
+                uri: url.pathname
+            };
+        }
+
+        if (fullDatasetCollectionData != null) {
+            switch (filterableDataTypeToUse) {
+                case filterableReportType.dataset:
+                case filterableReportType.datasetVersion:
+                    fullDatasetCollectionData.forEach((datasetCollectionItem) => {
+                        let singleReportItem = getMandatoryFilterableReportInformation(datasetCollectionItem);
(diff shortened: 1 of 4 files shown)


dist/legacy-assets/js/main.js
```diff
@@ -55782,15 +55782,47 @@ function viewReleaseSelector() {
 
 function viewReportDetails(collection, isPublished, $this) {
 
-    var details, reportDetails, published, events;
+    // Enum type of filterable data
```
dandwal: Use of Enum to avoid magic strings.

dist/legacy-assets/js/main.js
```diff
@@ -55782,15 +55782,47 @@ function viewReleaseSelector() {
...
+    const filterableReportType = {
+        'dataset': 0,
+        'datasetVersion': 1
+    };
+
+    // Retrieve all information for a specific type of filterable data report
+    function getFilterableReportData(filterableDataTypeToUse, fullDatasetCollectionData) {
```
dandwal: This function has been written with DRY and Open Close principles in mind.
Currently the only data used is the URI, however a usecase is present to expand this function and add Duration and size, some fields might be mandatory for all types of dataset reports whereas some might be specific to an edition, dataset, version or other.

The outer function `getFilterableReportData` is used for a specific type of filterable data report whereas the `getMandatoryFilterableReportInformation` is for mandatory data that is in all types of filterable reports.

dist/legacy-assets/js/main.js
```diff
@@ -55782,15 +55782,47 @@ function viewReleaseSelector() {
...
+        // Retrieve all mandatory information fields for filterable data report
+        function getMandatoryFilterableReportInformation(datasetCollectionItem) {
+            const url = new URL(datasetCollectionItem.uri);
+            return {
+                uri: url.pathname
+            };
+        }
+
+        if (fullDatasetCollectionData != null) {
+            switch (filterableDataTypeToUse) {
```
dandwal: This is a waterfalling switch used at current, however, this enables for specific data for a filterable data type to be added, and new types to append the switch.

PR #387, opened 2019-11-27
Upload journey select recipe page updated for new journey

dandwal:
### What

As part of the upload journey the select recipe page has now been reconstructed to look like the select a dataset page for the CMD version creation journey

### How to review

Check that a job is created and it progresses to the next screen - during development I noticed that at times this failed but I believe this is just an issue with my local stack and not the code base but verification would be useful.

### Who can review

Anyone except me


src/app/views/uploads/dataset/DatasetUploadsController.jsx
```diff
@@ -44,13 +41,14 @@ export class DatasetUploadsController extends Component {
         this.setState({ isFetchingData: true });
 
         const getRecipes = recipes.getAll();
-        const getJobs = datasetImport.getAll();
 
-        Promise.all([getRecipes, getJobs])
+        Promise.all([getRecipes])
```
jondewijones: Can we change this? Don't need Promise.all() for one promise

src/app/views/uploads/dataset/DatasetUploadsController.jsx
```diff
@@ -3,14 +3,14 @@ import { connect } from "react-redux";
 import { push } from "react-router-redux";
 import PropTypes from "prop-types";
 
-import { updateAllJobs, addNewJob, updateAllRecipes } from "../../../config/actions";
+import { addNewJob, updateAllRecipes } from "../../../config/actions";
 import recipes from "../../../utilities/api-clients/recipes";
 import datasetImport from "../../../utilities/api-clients/datasetImport";
 import notifications from "../../../utilities/notifications";
-import Definition from "../../../components/Definition";
+import log, { eventTypes } from "../../../utilities/log";
```
jondewijones: Can we change this to the new logging library? This is the old one

PR #398, opened 2020-01-28
Update accordion labels inline with website

jondewijones:
### What

Pass new label value to templates to give more descriptive headings. Some have been left as-is where there was no straightforward solution. They will need some extra work and should probably get looked at by UR/UX.

### How to review

Check new labels match their respective parts on the website

### Who can review

Anyone

### DIFF
--- a/src/legacy/js/functions/_externalLinkAccordionSection.js
+++ b/src/legacy/js/functions/_externalLinkAccordionSection.js
@@ -6,9 +6,21 @@
  * @param idField - HTML id for the template
  */
 
+function getExternalLinkAccordionSectionLabel(idField) {
+    switch(idField) {
+        case "filterable-dataset":
+            return "Try a new way to download data"
+        case "link":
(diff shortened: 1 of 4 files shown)


src/legacy/js/functions/_externalLinkAccordionSection.js
```diff
@@ -6,9 +6,21 @@
...
  */
 
+function getExternalLinkAccordionSectionLabel(idField) {
+    switch(idField) {
+        case "filterable-dataset":
+            return "Try a new way to download data"
+        case "link":
+            return "You might also be interested in"
+        default:
+          return `Related ${idField}`
```
daiLlew: Is there potential for the `idField` to be empty?

jondewijones: No. Well, very _very_ unlikely and worst case scenario it would return `Related undefined` rather than bork.

PR #401, opened 2020-02-11
Feature/health endpoint

DavidSubiros:
### What

- Replaced old healthcheck with dp-healthcheck
- Add checkers for Vault (if we're using encryption), and S3
- Migrated S3 functionality from upload.go to dp-s3 repository.
- Use go modules (with go version 1.13)
- Refactored Makefile/CI so that they don't depend on $GOPATH
- Refactored Makefile generate step - we only create a single generated file (either prod or debug) to prevent duplicated functions and variables. Fixed test for the debug case (created testBindataFileInfo).
- replaced go-ns/log with log.go

### How to review

- Verify that the changes make sense
- Go tests should pass in both debug and prod generated files:
  - `make node-modules` `make generate-go-debug` `make test-go`
  - `make node-modules` `make generate-go-prod` `make test-go`
- try to run Florence: `make debug` and try the /health endpoint: `curl GET 'http://localhost:8081/health'`. If you run again with `ENCRYPTION_DISABLED=true`, then you should only see the S3 Check in the list.

### Who can review

Anyone


Makefile
```diff
@@ -1,8 +1,11 @@
 BIND_ADDR ?= :8081
 BINPATH ?= build
 
+BUILD_TIME=$(shell date +%s)
+GIT_COMMIT=$(shell git rev-parse HEAD)
 VERSION=`git describe --tags`
```
nshumoogum: We do not need this variable, we should use `HC_VERSION` but rename to `VERSION`

main.go
```diff
@@ -35,89 +35,115 @@ var upgrader = websocket.Upgrader{}
 // Version is set by the make target
 var Version string
```
nshumoogum: Remove this variable

main.go
```diff
@@ -35,89 +35,115 @@ var upgrader = websocket.Upgrader{}
...
 
+var (
+	// BuildTime represents the time in which the service was built
+	BuildTime string
+	// GitCommit represents the commit (SHA-1) hash of the service that is running
+	GitCommit string
+	// HcVersion represents the version of the service that is running
+	HcVersion string
```
nshumoogum: Rename this variable to Version (that is now become available)

main.go
```diff
@@ -35,89 +35,115 @@ var upgrader = websocket.Upgrader{}
...
 		os.Exit(1)
 	}
 
-	zc := healthcheck.New(cfg.ZebedeeURL, "zebedee")
-	rtrc := healthcheck.New(cfg.RouterURL, "router")
-	dc := healthcheck.New(cfg.DatasetAPIURL, "dataset-api")
-	rc := healthcheck.New(cfg.RecipeAPIURL, "recipe-api")
-	ic := healthcheck.New(cfg.ImportAPIURL, "import-api")
+	// Create healthcheck object with versionInfo
+	versionInfo, err := healthcheck.NewVersionInfo(BuildTime, GitCommit, HcVersion)
```
nshumoogum: Any references to HcVersion should be updated to Version

Makefile
```diff
@@ -1,8 +1,11 @@
...
-LDFLAGS=-ldflags "-w -s -X main.Version=${VERSION}"
+HC_VERSION ?= $(shell git tag --points-at HEAD | grep ^v | head -n 1)
+LDFLAGS=-ldflags "-w -s -X 'main.Version=${VERSION}' -X 'main.HcVersion=${HC_VERSION}' -X 'main.BuildTime=$(BUILD_TIME)' -X 'main.GitCommit=$(GIT_COMMIT)'"
```
nshumoogum: Any references to `HC_VERSION` should be updated to `VERSION`

PR #413, opened 2020-03-18
Fix/latest dataset entry

rav-pradhan:
### What

`dp-dataset-api` was updated to return dataset instances in reverse order. The argument passed to `mapInstancesToState` in the controller has subsequently been reversed to align with this. 

This ultimately means that the "(latest)" text now appears beside the most recent dataset, rather than the oldest, when attaching new data to a collection in Florence

### How to review

- Log into Florence with `ENABLE_DATASET_IMPORT` flag set to `true`
- Create a collection
- Click on "Create/edit CMD dataset page"
- Select a dataset
- Select time-series edition
- Select "Create New Version"
- On this page check that the "(latest)" text appears next to the most recent dataset

### Who can review

Anyone but me

### DIFF
--- a/src/app/views/datasets-new/create/CreateVersionController.jsx
+++ b/src/app/views/datasets-new/create/CreateVersionController.jsx
@@ -165,9 +165,10 @@ export class CreateVersionController extends Component {
         return datasets
             .getCompletedInstancesForDataset(datasetID)
             .then(instances => {
+                const sortedInstancesByLastUpdated = instances.items.sort(this.sortByLastUpdated);
                 this.setState({
                     isFetchingInstances: false,
-                    instances: this.mapInstancesToState(instances.items.reverse())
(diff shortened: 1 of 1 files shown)


src/app/views/datasets-new/create/CreateVersionController.jsx
```diff
@@ -167,7 +167,7 @@ export class CreateVersionController extends Component {
             .then(instances => {
                 this.setState({
                     isFetchingInstances: false,
-                    instances: this.mapInstancesToState(instances.items)
+                    instances: this.mapInstancesToState(instances.items.reverse())
```
janderson2: Looks like it will work so I will approve, but I just had a thought.  It might be more resilient to add a sort to this so that it will always sort based on the date.  That way Florence is not dependent on any changes to the dataset API.

PR #421, opened 2020-04-17
Add vulnerability audit to CI build

jondewijones:
### What

Add vulnerability audit to CI build

### Who can review

Anyone


ci/audit.yml
```diff
@@ -0,0 +1,15 @@
+---
+platform: linux
+
+image_resource:
+  type: docker-image
+  source:
+    repository: onsdigital/dp-concourse-tools-nancy
```
janderson2, line 7: It may be better to add Nancy to the DP-concourse-tools-node-go Docker image so that we can still do npm audit when we get to that story.

ci/audit.yml
```diff
@@ -0,0 +1,15 @@
...
+    tag: latest
+
+inputs:
+  - name: florence
+    path: florence
```
janderson2, line 12: Technically not needed as it defaults to the input name.

PR #424, opened 2020-04-20
added sort class with static function to sort alphabetically

rav-pradhan:
### What

Datasets on the "Select a dataset" page weren't being ordered alphabetically. This PR fixes that by having a new `sort` class in the `utilities` directory. As there is another ticket that requires a list to be ordered alphabetically, it seemed better to create a common function for this logic.

### How to review

- Check the code changes make sense
- Go to the "Select a dataset" page from within a collection in Florence and ensure that the datasets are ordered alphabetically

### Who can review

Anyone but me

### DIFF
--- a/src/app/utilities/sort.js
+++ b/src/app/utilities/sort.js
@@ -1,4 +1,14 @@
 export default class sort {
+    /**
+     * A function to pass into a sort function's callback to order a list alphabetically.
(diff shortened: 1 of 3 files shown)


rav-pradhan:
Closing as upon discussion with Jon it was agreed that this logic should be within the publishing dataset controller itself.

PR #436, opened 2020-05-26
Add workflow preview component

jondewijones:
### What

Created a workflow preview component (name can change) that can work as preview for any content type (even datasets as i see that preview moving towards the design of this one) and set up to act as preview for the new homepage publishing journey for this PR.

### How to review

You can access the page directly by going to `/collections/{collection_id}/homepage/preview`

### Who can review

Anyone

### DIFF
--- a/src/app/views/workflow-preview/WorkflowPreview.jsx
+++ b/src/app/views/workflow-preview/WorkflowPreview.jsx
@@ -1,0 +1,91 @@
+import React, { Component } from "react";
+import { push } from "react-router-redux";
+import { Link } from "react-router";
+import PropTypes from "prop-types";
+
+import url from "../../utilities/url";
+import notifications from "../../utilities/notifications";
+
+import Iframe from "../../components/iframe/Iframe";
+
+const propTypes = {
+    location: PropTypes.PropTypes.shape({
+        pathname: PropTypes.string
+    }),
+    params: PropTypes.shape({
+        collectionID: PropTypes.string.isRequired
+    }),
+    dispatch: PropTypes.func.isRequired
+};
+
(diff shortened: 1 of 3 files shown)


src/app/views/workflow-preview/WorkflowPreview.jsx
```diff
@@ -0,0 +1,91 @@
...
+        const previousUrl = url.resolve("../");
+        this.props.dispatch(push(previousUrl));
+    };
+
+    // getPreviewIframeURL returns the url for the content to be reviewed when the url is
+    // in the format: /florence/collections/collection-id/{content/to/preview/url}/preview
+    // returns '/' incase the url is '/homepage' or errors
+    getPreviewIframeURL = path => {
+        try {
+            const regex = /(?:\/[^/]+){3}(?<contentURL>.+)\/preview/;
```
rav-pradhan, line 36: Wonder if we could give this a more meaningful name? Regex has never been a strong point of mine, so having a name that conveys what this pattern actually corresponds to could be helpful - possibly in particular when glancing through the code in the future

src/app/views/workflow-preview/WorkflowPreview.jsx
```diff
@@ -0,0 +1,91 @@
...
+                isDismissable: true
+            });
+            return "/";
+        }
+    };
+
+    render() {
+        this.getPreviewIframeURL(this.props.location.pathname);
+        return (
+            <div>
```
rav-pradhan, line 57: Is it worth updating to use more semantic markup across the render method? It's something I need to be more mindful of when writing my components, too

PR #440, opened 2020-07-14
Feature/edit homepage

rav-pradhan:
### What

Implemented the Edit Homepage functionality. Owing to the size of the feature, this PR was previously broken down into other PRs that have all been approved and merged in here. In theory, then, this PR should only need a sense check 

This PR includes:

- Edit homepage form/modal
- Preview functionality

### How to review

Check that the code changes align with what had previously been reviewed
May be worth doing a test in the UI as well just to make sure all is working as intended

### Who can review

Anyone but me


src/app/views/homepage/edit/EditHomepageController.jsx
```diff
@@ -0,0 +1,362 @@
...
+    handleOnSaveError = message => {
+        notifications.add({
+            type: "warning",
+            message: `${message}. You can try again by pressing save.`,
+            isDismissable: true
+        });
+    };
+
+    redirectTo = route => {
+        window.location = window.location.origin + route;
```
jondewijones, line 315: Is there a reason we're using `window.location` instead of router? e.g. `this.props.dispatch(push(uri))` ?

rav-pradhan: Have made suggested change - tested and looks to still work as intended too

PR #443, opened 2020-07-17
Fix/publishing journey collection

rav-pradhan:
### What

Fixes for publishing journey:

- Added defaults and a check to the collection state code to prevent catch block code running every time a Collection is loaded without any events for the homepage
- Typo fix in preview toast message
- Extracted method refactor for navbar. The navbar will now show the collection being worked on when on the Edit Homepage screen

### How to review

Check that the code makes sense and that the changes described above are reflected

### Who can review

Anyone but me

### DIFF
--- a/src/app/views/homepage/edit/EditHomepageController.jsx
+++ b/src/app/views/homepage/edit/EditHomepageController.jsx
@@ -91,25 +91,30 @@ export class EditHomepageController extends Component {
         collections
             .getContentCollectionDetails(this.props.params.collectionID)
             .then(collection => {
-                const lastEventIndex = collection.eventsByUri["/data.json"].length - 1;
-                const lastEvent = collection.eventsByUri["/data.json"][lastEventIndex];
-                let collectionState = "";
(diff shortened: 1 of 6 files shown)


src/app/views/homepage/edit/EditHomepageController.jsx
```diff
@@ -91,25 +91,30 @@ export class EditHomepageController extends Component {
...
-                    case "COMPLETED":
-                        collectionState = "complete";
-                        break;
-                    case "EDITED":
-                        collectionState = "inProgress";
-                        break;
-                    default:
-                        collectionState = "reviewed";
+                const pageHasEvent = collection.hasOwnProperty("eventsByUri") && collection.eventsByUri.hasOwnProperty("/data.json");
+                if (pageHasEvent) {
```
jondewijones, line 95: In the case there is no page event what would happen? Should we set a default state? Throw an error and prevent editing and instruct user to refresh? Only allow save and preview? 

Happy to discuss

rav-pradhan: Have updated the form disabling logic so that it is disabled by default. The form is then enabled if all data is successfully fetched

PR #450, opened 2020-08-12
Feature/proxy through api router

DavidSubiros:
### What

- Added health checker for router API
- Refactored code:
  - `main` only creates service, with dependency list, and waits for fatal error or signal
  - created a new `service` package, with run/graceful shutdown, register checker, and router creation
  - handlers and directors in service package, in their respective files,
  - split and renamed tests according to the new code location.
  - upgraded all dependencies to use dp-net ContextKey and remove old go-ns dependencies.
- Modified directors to use to dp-router-pai URL:
  - Dataset API
  - Recipe API
  - Import API
- these directors now append dp-router-api version to path (as expected)
- Modified configuration according to new proxying strategy:
  - Removed `AatasetAPIURL`, `RecipeAPIURL`, `ImportAPIURL`
  - Added `APIRouterURL`, `APIRouterVersion`
- Extended directors unit tests to validate that the api version is correctly prepended, and the prefix is trimmed, as before.

### How to review

- Make sure code changes make sense
- Unit tests should pass

I tested locally by trying to upload a dataset, with an api-router running with auditing, and checked that audit events were correctly generated. I needed to remove CORS middleware, but once merged to develop, it should pass the CORS validation.

### Who can review

Anyone.


DavidSubiros:
> Changes look good. I have done a bunch of local testing in different areas of florence:
> 
> * browse website via florence
> * make changes to website content
> * review and approval of content
> * publish
> * import dataset
> * filter journey in preview
> * users / teams
> 
> We should consider not auditing for "Path":"/clickEventLog". There are many of these that get generated when using Florence, and typically have an equivalent entry for API access.

I can create a branch in dp-api-router to ignore `clickEventLog` in the auditing middleware (similarly to what we did for `ping`)

PR #457, opened 2020-09-07
Add image upload for homepage tiles

jondewijones:
### What

Add image upload for homepage tiles

### How to review

Speak to me :)

### Who can review

Anyone


src/app/views/homepage/edit/EditHomepageItem.jsx
```diff
@@ -41,20 +94,268 @@ export default class EditHomepageItem extends Component {
...
+                url: "",
+                title: "",
+                altText: ""
+            }
+        });
+        await this.createImageRecord();
+        this.bindInput();
+    };
+
+    renderImagePreview = () => {
```
rav-pradhan, line 274: I think we should consider refactoring this method as it has multiple if statements. The various import/image fetching booleans only seem to be used in this method. Correct me if I'm wrong though

Could we maybe store the strings we return here as constants, and then in the various `setState` calls where the booleans are being set, we could instead have a single variable that is manipulated. Something like: `setState({ currentImageImportStatus: IMPORTING_IMAGE })`? 

That means in this method, we could then have a single if statement to check the image data object is there, and then a single `return <p>{this.state.currentImageStatus}</p>` beneath it. Let me know what you think

src/app/views/homepage/edit/EditHomepageItem.test.js
```diff
@@ -70,3 +140,135 @@ describe("event handlers", () => {
...
+    });
+
+    it("errors cause notification", async () => {
+        image.update.mockImplementationOnce(() => Promise.reject({ status: 404 }));
+        await wrapper.instance().addUploadToImageRecord(mockImage.id, mockImage.upload.path);
+        expect(mockNotifications.length).toBe(1);
+    });
+});
+
+describe("get image upload ", () => {
```
rav-pradhan, line 204: If we did change up the way we're storing the current image/import status, we'd need to update these tests too. `expect(wrapper.state("currentImageImportStatus").toEqual("Image is loading")` etc

rav-pradhan (commented):
Appreciate there are still some bits that are WIP off the back of the image api/import service work. So this isn't a full review as I've not yet managed to test it end-to-end, but have had a first pass through the code. Couple comments, let me know what you think

rav-pradhan (approved):

rav-pradhan (approved):

rav-pradhan (approved):

PR #465, opened 2020-10-06
Feature/add dp upload service config

sandypadmanabhan:
### What

Added dp-upload-service to config
Added feature flag "ENABLE_UPLOAD_SERVICE" to shared Config
Added routes to handle the upload service endpoints 

### How to review
check config for
UploadServiceURL:           "http://localhost:25100"
and "ENABLE_UPLOAD_SERVICE" feature flag
Check route added is correct
Check unit tests are added

### Who can review
Anyone


config/config.go
```diff
@@ -14,6 +14,7 @@ type Config struct {
 	RouterURL                  string        `envconfig:"ROUTER_URL"`
 	DatasetControllerURL       string        `envconfig:"DATASET_CONTROLLER_URL"`
 	TableRendererURL           string        `envconfig:"TABLE_RENDERER_URL"`
+	UploadServiceURL           string        `envconfig:"UPLOAD_SERVICE_URL"`
```
DavidSubiros, line 17: Do we need to proxy upload service calls to the API router? If we do, then we should not have this config (and use the API router URL instead)

config/config.go
```diff
@@ -47,9 +49,10 @@ func Get() (*Config, error) {
 		RouterURL:                  "http://localhost:20000", // Frontend router
 		DatasetControllerURL:       "http://localhost:24000",
 		TableRendererURL:           "http://localhost:23300",
+		UploadServiceURL:           "http://localhost:25100",
```
DavidSubiros, line 52: same comment as above

service/directors.go
```diff
@@ -31,6 +31,13 @@ func recipeAPIDirector(apiRouterVersion string) func(req *http.Request) {
 	}
 }
 
+func uploadServiceAPIDirector(apiRouterVersion string) func(req *http.Request) {
+	return func(req *http.Request) {
+		director(req)
+		req.URL.Path = fmt.Sprintf("/%s%s", apiRouterVersion, req.URL.Path)
```
DavidSubiros, line 37: the apiRouterVersion is required only if we proxy calls to the API router. 
If we don't do it, then you can ignore the apiRouterVersion, and just use the base director (without modifying the URL path)

service/directors_test.go
```diff
@@ -67,4 +67,11 @@ func TestDirectors(t *testing.T) {
 		datasetAPIDirector("v123")(request)
 		So(request.URL.String(), ShouldEqual, "/v123/foo")
 	})
+
+	Convey("UploadService API proxy director function appends the provided router api version to the provided path' from the request URL", t, func() {
```
DavidSubiros, line 71: same comment as above (this is true if we proxy calls through API router)

DavidSubiros (commented):
a few comments about API router prefix.
Do we need to proxy upload service calls through API router?

DavidSubiros (approved):
Approved, but before enabling this feature, we will probably need to add the proper handler in the API router. (I guess we need another Trello card to do that)
Route to proxy in API proxy:
```
router.Handle("/upload{uri:.*}", uploadServiceAPIProxy)
```

PR #468, opened 2020-10-16
Build URL for the latest version instead of using API response URL

CarlHembrough:
### What
Using the dataset API response URL contains the `v1` prefix as it goes through the API router. Using this path causes the subsequent request to fail. Building the URL ensures that this doesn't happen, and it more consistent with the rest of the code.

This bug surfaced when trying to add a dataset to a collection. Once on the dataset metadata edit screen an error would show as the latest version could not be found. Looking at the URL that it's trying to use, you can see the `/v1` in the path: `http://localhost:8081/dataset/v1/datasets/cpih01/editions/time-series/versions/2`

### How to review
Review changes - check that I haven't broken anything

### Who can review
Anyone familiar with the Florence JS code

### DIFF
--- a/src/app/utilities/api-clients/datasets.js
+++ b/src/app/utilities/api-clients/datasets.js
@@ -96,8 +96,10 @@ export default class datasets {
                 return;
             }
 
-            const latestVersionURL = url.resolve(dataset.current.links.latest_version.href);
-            const latestVersion = await http.get(`/dataset${latestVersionURL}`).catch(error => reject(error));
+            const latestVersionPath = dataset.current.links.latest_version.href.replace(dataset.current.links.self.href, "");
+            const latestVersionURL = datasetURL + latestVersionPath;
+
+            const latestVersion = await http.get(latestVersionURL).catch(error => reject(error));
             resolve(latestVersion);
         });
     }
--- a/src/package-lock.json
+++ b/src/package-lock.json
@@ -64,9 +64,9 @@
           },
           "dependencies": {
             "lodash": {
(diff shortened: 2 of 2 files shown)


src/app/utilities/api-clients/datasets.js
```diff
@@ -96,8 +96,10 @@ export default class datasets {
                 return;
             }
 
-            const latestVersionURL = url.resolve(dataset.current.links.latest_version.href);
-            const latestVersion = await http.get(`/dataset${latestVersionURL}`).catch(error => reject(error));
+            const latestVersionID = dataset.current.links.latest_version.id;
```
nshumoogum, line 99: Is it possible that the latest version when calling GET `/dataset/{id}` may return a version for a different edition of the dataset than the one you care about?

Where there are two or more editions to a dataset?

I guess it depends if this method is only used from a dataset page.

nshumoogum: Should you be calling the edition endpoint? And getting latest version from there instead?

CarlHembrough: A good point - I think it is possible for a version of a different edition to be returned on the dataset endpoint. I guess it's a product question whether we should pull data from the latest version, or the latest version of that edition 🤔

nshumoogum: I think there could be a possibility to create two distinct methods in code, and both would have use cases on the website, depending on where what functionality you are trying to provide.

nshumoogum (approved):

PR #472, opened 2020-10-27
Fix/visualisations preview

rav-pradhan:
### What

There would be a bug where an overlay would applied to the preview screen in the old workspace, when moving from a visualisations page to a non-vis one in the browse tree. This PR fixes that and also updates the logic so that users will be able to go through the various html files in a visualisation upload from the browse tree rather than having to go to the Edit Visualisations page

The implementation specifically is as follows:

- A new conditional in the `treeNodeSelect` function checks to see whether the user has clicked on a visualisation in the browse tree, determined by the value of the `data-url` attribute
- If it is a visualisation selected, then a new function is called that fetches the visualisation data, which then in turn calls the `visualisationEditor` function that renders the `<select>` element in lieu of the URL field on the preview screen
- The logic in this function has been tweaked slightly to ensure that the first html file in the visualisation is selected by default, too. Previously, visualisations would look incomplete / improperly loaded when viewing from the browse screen
- Code related to the overlay has been removed as it no longer seems to serve a purpose

#### Before

A visualisation page on the browse screen:
image

Going to a non-vis page after being on the Edit Vis screen:
image

#### After

Visualisations can now be previewed from the browse tree (URL changes to a select dropdown)
image

Navigating to a non-vis page afterwards no longer has the obscuring overlay
image

### How to review

- Ensure code changes make sense
- I extracted some logic into its own function for reuse. I've been able to upload a visualisation and approve a collection with a mix of visualisations and non-vis pages, and that all works as intended. It may be worthwhile having another set of eyes on this, though

### Who can review

Anyone familiar with Florence

### DIFF
--- a/src/legacy/js/functions/_createWorkspace.js
+++ b/src/legacy/js/functions/_createWorkspace.js
@@ -342,18 +342,6 @@ function updateBrowserURL(url) {
         url = Florence.globalVars.pagePath;
     }
     $('.browser-location').val(Florence.babbageBaseUrl + url);
-
-    // Disable preview for visualisations
-    var isVisualisation = url.split('/')[1] === "visualisations";
-    if (isVisualisation && $('#browse.selected').length > 0) {
-        $('.browser').addClass('disabled');
-        return;
-    }
-
-    // Enable the preview if we're viewing a normal page and the preview is currently disabled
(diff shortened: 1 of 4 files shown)


justinpjose (approved):
Went through the journey with Ravi and it works as expected. Nicely done Ravi

PR #477, opened 2020-11-25
Fix/remove old table builder

NoelleDL:
### What

Removed buttons and code to generate tables with the old table builder

### How to review

On develop, see that there are two buttons to generate tables ("Add table") and ("Add table (new)") when editing a collection (within the Table & Content accordions). Pull this branch, see that there is now one button that links to the new table builder. All references to the old code are gone. 

### Who can review

Anyone but me.


nshumoogum (approved):
Code looks good to me, ran florence using make dev (to update dist and assets) and can see changes look good; old table builder removed

PR #478, opened 2020-12-04
CMD Uploads now say when they are completed

dandwal:
### What

When CMD upload is complete it will now update the page automatically and no longer require constant refreshes to find out if the upload is complete

### How to review

Upload a dataset, ensure that the page updates once complete to notify you that it has completed.
Check the network tab to ensure after completion there are no more requests going to `dataset/instances/{id}`

### Who can review

Anyone except me


rav-pradhan (approved):
Have tested locally and can see the changes as described. Code looks good too, setState callback is a good touch - nice work. Only question: should we add a test suite for this?

PR #482, opened 2020-12-16
Fix publishing queue collections details failures

jondewijones:
### What

Wait for all requests to complete even if there's failures. Capture fails and present them back to user

### How to review

- Check the publishing queues works as normal when no errors
- When there are errors the publishing queue should continue with work with the error collection accordion displaying the error and others working as normal

To simulate requests error you can edit the `getCollectionDetails` function so that certain errors fail. E.g. on a certain collection ID change the url so it will 404

### Who can review

Anyone

### DIFF
--- a/src/legacy/templates/publishDetails.handlebars
+++ b/src/legacy/templates/publishDetails.handlebars
@@ -8,67 +8,72 @@
         {{#each collectionDetails}}
             <div class="accordion js-accordion">
                 <div class="accordion__title js-accordion__title">
-                    <h3 class="collection-name" data-id="{{id}}">{{name}}</h3>
+                    <h3 class="collection-name" data-id="{{id}}">{{#if name}}{{name}}{{else}}{{id}}{{/if}}</h3>
                 </div>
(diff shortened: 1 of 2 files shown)


src/legacy/templates/publishDetails.handlebars
```diff
@@ -8,67 +8,72 @@
...
-                            {{#if pageDetails}}
-                                {{#each pageDetails}}
-                                    <li class="page-list__item"><span class="page__item page__item--{{type}}"
-                                              data-path="{{uri}}">{{description.title}}{{#if description.edition}}
-                                        : {{description.edition}}{{/if}}</span>
-                                        <div class="page__buttons page__buttons--list">
-                                            <button class="btn btn--warning btn-page-delete">Remove from this publish</button>
-                                        </div>
+                    {{#if error}} 
+                        <p>This collection details has failed to load.</p>
```
janderson2, line 19: It might read better to change this to:

```suggestion
                        <p>This collection's details have failed to load.</p>
```

janderson2 (approved):

PR #488, opened 2021-02-08
Feature/use controller for edit dataset metadata

jondewijones:
### What

Update edit dataset metadata component to use `dp-publishing-dataset-controller` rather than interface directly with APIs.

- Update mapping logic (mostly unchanged state to reduce amount of changing parts in PR)
- Add test coverage for new methods (out of scope to add it for all funcs)
- Remove old/unused code

### How to review

- Code review
- Test review - pull latest `dp-publishing-dataset-controller` and try publish a new version. Also worth checking some edge cases around trying to edit a version while it's in another collection, etc.

### Who can review

Anyone


src/audit-whitelist.json
```diff
@@ -0,0 +1,11 @@
+{
+    "ignore": [
+        { "id": "e7bea77b-f37f-4290-a069-08fc9c4b6ac0", "package": "acorn" },
```
janderson2, line 3: Have security vulnerability tickets been raised in the triage column on the backlog for this? If not, could we please raise those before merging this so we don't forget to come back to these.

src/audit-whitelist.json
```diff
@@ -0,0 +1,11 @@
+{
```
nshumoogum, line 1: If the name of the file does not matter I think we should start thinking about replacing the use of `whitelist` with `allowedlist`

janderson2: Good shout. I would suggest potentially `allowlist` rather than `allowedlist` to be consist with what I have seen elsewhere?

jondewijones: Used "whitelist" as that was the option name but i prefer "allowlist" 👍

src/audit-whitelist.json
nshumoogum: Either way i have added the missing stories to the backlog to cover all these low risk vulnerabilities

nshumoogum (approved):

PR #492, opened 2021-02-23
Remove duplicate preview components => use WorkFlowPreview instead of DatasetPreview (CMD) component

jondewijones:
### What

- Remove duplicate preview components => use WorkFlowPreview instead of DatasetPreview (CMD) component

### How to review

Preview a CMD dataset in the publishing journey... it should continue to work as normal 

### Who can review

Anyone


rav-pradhan (approved):
Checked locally and could still preview dataset

PR #504, opened 2021-04-06
Fix/cmd edit modal issue

CarlHembrough:
### What
Allow the usage notes of a dataset version to be updated. Before this change an error would occur when trying to add a usage note.

### How to review
Review changes / tests

### Who can review
Anyone, but probably @jondewijones is most familiar with this area


AngelaLW:
I know this isn't really related to usage notes, but as you've fixed all the modal popups on the dataset edit page I've noticed the `relatedDatasets` object in the files you've changed has `description` as an property in the modal, but in the Dataset API swagger docs the related datasets look like this:
`"related_datasets": [
    {
      "href": "string",
      "title": "string"
    }
  ],`

Also the latestChanges object looks like this:
`"latest_changes": [
    {
      "description": "string",
      "name": "changes in classification",
      "type": "summary of changes"
    }
  ],` which is also slightly different to the fields defined in the Florence side....

CarlHembrough:
> I know this isn't really related to usage notes, but as you've fixed all the modal popups on the dataset edit page I've noticed the `relatedDatasets` object in the files you've changed has `description` as an property in the modal, but in the Dataset API swagger docs the related datasets look like this:
> `"related_datasets": [ { "href": "string", "title": "string" } ],`
> 
> Also the latestChanges object looks like this:
> `"latest_changes": [ { "description": "string", "name": "changes in classification", "type": "summary of changes" } ],` which is also slightly different to the fields defined in the Florence side....

Good points, though I think both cases should be analysed as separate tasks. It looks like both cases require API changes, and it should be checked that these fields are actually used. 

The related datasets are being passed to the API, but not being persisted. The latest changes field is also passed to the API but not persisted, as well as the fields not matching up.

nshumoogum (approved):

nshumoogum:
> > I know this isn't really related to usage notes, but as you've fixed all the modal popups on the dataset edit page I've noticed the `relatedDatasets` object in the files you've changed has `description` as an property in the modal, but in the Dataset API swagger docs the related datasets look like this:
> > `"related_datasets": [ { "href": "string", "title": "string" } ],`
> > Also the latestChanges object looks like this:
> > `"latest_changes": [ { "description": "string", "name": "changes in classification", "type": "summary of changes" } ],` which is also slightly different to the fields defined in the Florence side....
> 
> Good points, though I think both cases should be analysed as separate tasks. It looks like both cases require API changes, and it should be checked that these fields are actually used.
> 
> The related datasets are being passed to the API, but not being persisted. The latest changes field is also passed to the API but not persisted, as well as the fields not matching up.

Re. Fields not aligning in Dataset API and persisting to database, I believe they do and they are. The swagger spec is incorrect re. related_datasets fields which does contain a description.

The mismatch in field name for latest_changes => name (in API) and title (in florence).

The reason why you may think this is not persisting to the database is because the data is not being passed to the API, Florence sends the data to a frontend controller that then sends the data onto the API. The frontend controller has it's own models based on the library dp-api-clients-go which is missing the description field for related_datasets. The latest_changes expects to see name not title in the controller.

So stories needed to fix these would be:

1) Update dataset API swagger spec to include `description` for a related_dataset (API is now correct)
2) Update dp-api-clients-go to include `description` in it's model of a dataset document
3) Import new version of library into frontend controller (assuming dp-frontend-dataset-controller)
4) Update florence to send the `title` field to controller as `name` instead so it can be decoded in the controller

jondewijones (approved):

PR #507, opened 2021-04-08
New Auth Page

dandwal:
⚠️ Looks bigger than it actually is! ⚠️ 

### What

We are making a new Authentication user journey
This is the first PR of this journey which is the beginning of laying out the ground work

The idea is to keep the original Auth journeys and develop the New Auth journey in parallel. We will use a feature flag to toggle between the two different journeys. Once the entire Auth Epic work is complete we will delete the old login and auth work then remove the feature flags so that the new journeys are all that are left.

A decision has been made in the interim that if a change is minor and adds very little work we can make the changes without the feature flag and apply it to the current journey. An example of this is the Login screen title - which in this PR is applied to the new Sign In screens and the old login screen.

The terms:
- verb 'Log in'
- noun 'Login'
Will be replaced in upcoming work in one batch so for the time being from the users perspective they will see the term log in until the change of term work is added to the sprint.

The spirit of this Auth work is to keep PRs and tasks small. So rather than completing a whole page at a time, little changes will be introduced on each PR

#### So what is actually in this PR

- New feature flag used "ENABLE_NEW_SIGN_IN"
- 'Login' Florence page title replaced with "Log in to your Florence account"
- Line-height on the title increased
- Duplication of the loginController, LoginForm and LoginController.test - these will look more and more different in future PRs and tidied up a bit
- Tests (Config and LoginController) updated (but to be improved in future)

### How to review

1. Run Florence without the feature flag
2. Follow the login journey and ensure everything works like normal
3. The title of that login page should read "Log in to your Florence account"
4. Stop Florence and run: `make node-modules-react; make debug ENABLE_DATASET_IMPORT=true ENCRYPTION_DISABLED=true ENABLE_NEW_SIGN_IN=true;`
5. Navigate to the Florence login screen http://localhost:8081/florence/login
6. Ensure that the login screen title now says "Log in to your Florence account".

### Who can review

Anyone except me


src/app/views/login/SignInController.jsx
```diff
@@ -0,0 +1,244 @@
+import React, { Component } from "react";
```
dandwal, line 1: As per the description, this is a duplicate file that will be edited in future PRs
- line 6 is however different

src/app/views/login/SignInController.test.js
```diff
@@ -0,0 +1,60 @@
...
+});
+
+test("Check redirect doesn't work if not authenticated", () => {
+    const props = {
+        dispatch: function() {},
+        rootPath: "/florence",
+        isAuthenticated: false
+    };
+    const component = mount(<LoginController {...props} />);
+    expect(component.find("h1").text()).toBe("Log in to your Florence account");
```
dandwal, line 47: As per the description, this is mostly a duplicate file that will be worked on in future PRs.
- This line however is changed.

src/app/views/login/SignInForm.jsx
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    render() {
+        const inputs = this.props.formData.inputs;
+        const isSubmitting = this.props.formData.isSubmitting;
+
+        return (
+            <div className="grid grid--justify-center">
+                <div className="grid__col-3">
+                    <h1 className="line-height--46">Log in to your Florence account</h1>
```
dandwal, line 34: As per the description, this is mostly a duplicate file that will be worked on in future PRs.

- This line however is changed.

src/app/views/login/LoginForm.jsx
```diff
@@ -31,7 +31,7 @@ export default class LoginForm extends Component {
         return (
             <div className="grid grid--justify-center">
                 <div className="grid__col-3">
-                    <h1 className="text-center">Login</h1>
```
dandwal, line 34: Old login page update with new page title and styling.

src/app/views/login/SignInController.test.js
```diff
@@ -0,0 +1,60 @@
...
+});
+
+test("Does password change form appear on state change", () => {
+    const props = {
+        dispatch: function() {},
+        rootPath: "/florence",
+        isAuthenticated: false
+    };
+    const component = mount(<LoginController {...props} />);
+    expect(component.find(".modal__overlay").length).toBe(0);
```
hamishtaplin, line 57: Is this how tests are generally written in Florence? Asserting on a class that's used for styling couples your test to the styling mechanism — eg. if the class name was changed due to a refactor it would break this unrelated test, making it fragile.

It's better to use selectors that are either:

- something the user will see (eg. title of the modal)
- explicitly for tests (eg. a `data-testid="modal"`) attribute

This is a good article on this: https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change

src/app/views/login/SignInController.jsx
```diff
@@ -0,0 +1,244 @@
...
+        this.setState({ isSubmitting: true });
+
+        this.handleLogin(credentials);
+    }
+
+    handleInputChange(event) {
+        const id = event.target.id;
+        const value = event.target.value;
+
+        switch (id) {
```
hamishtaplin, line 153: I know this is a copy/paste of another file but you should always have a `default` case on a switch that throws an error if the other cases don't match—this gives you a bit of a defence against typos, etc

src/app/views/login/LoginForm.jsx
```diff
@@ -31,7 +31,7 @@ export default class LoginForm extends Component {
...
+                    <h1 className="line-height--46">Log in to your Florence account</h1>
```
hamishtaplin, line 34: Surely you want to make this change more globally? The page title should be consistent across all pages.

src/app/views/login/SignInController.test.js
dandwal: This was an interesting article and think will implement in one of the other PRs as discussed in standup.

src/app/views/login/SignInController.jsx
dandwal: 100% agree on this and will add to one of the next PRs for the login work.

src/app/views/login/LoginForm.jsx
dandwal: Great line-height info. I have made this relative to the font size now and across all H1s I have checked numerous screens that use H1s in Florence and they all look fine.

src/app/views/login/SignInForm.jsx
```diff
@@ -0,0 +1,53 @@
...
+                onChange: PropTypes.func,
+                error: PropTypes.string
+            })
+        ),
+        onSubmit: PropTypes.func,
+        isSubmitting: PropTypes.bool
+    })
+};
+
+export default class LoginForm extends Component {
```
rav-pradhan, line 22: As we're only really working with props and presentational elements here, this could be a good candidate for refactoring into a functional component. something like:

```js
// or export default ({ // }) => etc.
const LoginForm = ({ inputs, isSubmitting, onSubmit }) => {
    return (
        // markup
    )
}
```

This is quite a specific component so won't be overly reused I suspect, but I think destructuring the props would make it clearer as well what exactly is required, and also reduce the need to set new variables to abstract away the `this.props.formData` reference

rav-pradhan (commented):
I can see that the feature flag is working and that the actual change as described has been made to the sign in form

Appreciate a lot of this has just been ported over to start off with for now, but I've left a couple comments in any case - let me know if you're going to address these now or in a later PR. Happy to approve once clarified 👍🏽

rav-pradhan (approved):
Tested locally, looks good, and refactors will be taken on in future pieces of work as discussed 👍🏽

PR #510, opened 2021-04-12
Feature/305 301 fix vulnerabilities

Marianasilvago:
### What

Updated dependency versions of auditjs and added resolutions for ini.

### How to review

Check the dependencies and validate the build is passing

### Who can review

Anyone with understanding of npm dependencies.


src/package.json
```diff
@@ -109,6 +109,7 @@
   "resolutions": {
     "lodash": "^4.17.15",
     "lodash-es": "^4.17.15",
-    "js-yaml": "^3.13.1"
+    "js-yaml": "^3.13.1",
+    "ini": "^1.3.6"
```
nshumoogum, line 113: Would it be better to go to version 1.3.8?

nshumoogum (approved):

PR #511, opened 2021-04-14
Feature/305 node fetch moment version

Marianasilvago:
### What

Update multiple dependency version to fix vulnerabilities.

### How to review

Check the passing tests and the dependency versions

### Who can review

Anybody with understanding of florence and nodejs.


src/package.json
```diff
@@ -107,9 +107,11 @@
...
   },
   "resolutions": {
-    "lodash": "^4.17.15",
-    "lodash-es": "^4.17.15",
+    "lodash": "^4.17.21",
+    "lodash-es": "^4.17.21",
     "js-yaml": "^3.13.1",
-    "ini": "^1.3.8"
+    "ini": "^1.3.8",
+    "moment": "^2.19.3",
```
s-xton, line 114: could we move to 2.29.1 here ?

src/package.json
```diff
@@ -87,12 +87,12 @@
...
+    "minimongo": "^6.8.2",
     "node-sass": "^5.0.0",
     "onchange": "^6.1.0",
     "path": "^0.12.7",
-    "react": "^15.6.2",
-    "react-dom": "^15.6.2",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
     "react-redux": "^5.1.2",
     "react-router": "^3.2.6",
```
s-xton, line 97: can you confirm that the non-upgraded versions of react are compatible with v17

Marianasilvago: React router current  version is 3.2.6

React router latest version 6.0.0 depends on React 16.x only. 
ref: https://github.com/ReactTraining/react-router/blob/master/package.json#L39

Current codebase would have to evolve a lot to support  react v17.x and react-dom v17.x
Hence falling back to v15.x

s-xton (approved):

PR #514, opened 2021-04-19
Fix/alt text not saved

NoelleDL:
### What

User was unable to input text into alt text and title fields when uploading an image. 

### How to review

In Florence try to type in the alt text field when uploading an image and see that changes aren't saved. Pull this branch, type in alt text field and see that changes are saved. 

### Who can review

Anyone but me.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3 @@
+{
+  "lockfileVersion": 1
+}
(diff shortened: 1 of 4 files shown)


NoelleDL:
Closing, didn't proceed with this fix

PR #516, opened 2021-04-27
Update versions list page to use publishing-dataset-controller

jondewijones:
### What

Update versions list page to use publishing-dataset-controller

### Who can review

Anyone


src/app/views/datasets-new/versions/DatasetVersionsController.jsx
```diff
@@ -39,27 +36,28 @@ export class DatasetVersionsController extends Component {
...
-    getDataset = datasetID => {
-        this.setState({ isFetchingDataset: true });
+    getAllVersions = (datasetID, editions) => {
+        this.setState({ isFetchingData: true });
         return datasets
-            .get(datasetID)
-            .then(dataset => {
+            .getVersionsList(datasetID, editions)
+            .then(response => {
+                console.log(response);
```
rav-pradhan, line 46: We should remove this log

rav-pradhan (commented):
Looks good, tested locally as well and seemed to run as before. Just a couple tiny comments

src/app/views/datasets-new/versions/DatasetVersionsController.jsx
jondewijones: We should and have :)

src/app/utilities/api-clients/datasets.js
```diff
@@ -210,4 +210,10 @@ export default class datasets {
             return response;
         });
     }
+
+    static getVersionsList(datasetID, editionID) {
+        return http.get(`${datasetControllerURL}/datasets/${datasetID}/editions/${editionID}/versions`).then(response => {
```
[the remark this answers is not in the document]
jondewijones: Agreed. Clearer is always better. I've changed the variable name in this file for now. I think to change it through the codebase would be a separate task

rav-pradhan (approved):

PR #517, opened 2021-05-04
New Show / Hide password pattern

dandwal:
### What

Instead of showing the show/hide in the password input field on the login screen, it is now shown as a checkbox under the password field.
- Focus styling updated
- Refactored input component a little (once all auth work has been complete it can undergo a final refactor to remove the option of the show/hide from within the input completely).

Note: this work is behind a feature flag: ENABLE_NEW_SIGN_IN

### How to review

Run florence like so: 
```bash
make debug ENABLE_DATASET_IMPORT=true ENCRYPTION_DISABLED=true ENABLE_NEW_SIGN_IN=true
```
Ensure that the feature flag ENABLE_NEW_SIGN_IN is true. 
- Test that the login show password field works
- Test that other password fields still show the old show/hide (e.g. add a new user, change password etc)

Stop and restart Florence this time with the feature flag disabled ENABLE_NEW_SIGN_IN=false 
- The focus styles should show the new styles but the show/hide should remain like they do on prod and develop.

### Who can review

Anyone except me


sandypadmanabhan (approved):
LGTM

PR #523, opened 2021-05-12
Make the failed upload of dataset title clearer

dandwal:
### What

User made a publishing support request due to an error appearing and was unsure why it was appearing. This PR makes it clear why the error is being thrown and should enable the user to realise what they need to do to resolve the issue. Without dev support.

New error:
Screenshot 2021-05-12 at 09 34 34

### How to review

Check that only the error message has changed. Ensure that the Error message is clear.

### Who can review

Anyone except me


eldeal (commented):
Many of our edition titles are years, so are exactly 4. Curious how that is getting past the validation, as I was expecting the rule to be 'at least 4'

eldeal (approved):

PR #528, opened 2021-05-19
Panel component added

dandwal:
### What

As part of the 'authentication' and 'authorisation' work, a new panel component is required in Florence.
Rather than having a massive PR for 'Sign in Validation' work, I decided to split this component out into its own PR - so you won't see it in use quite yet!

This component should meet our typical accessibility requirements.
This has been tested with: 
- Mac VoiceOver
- Keyboard navigation
- New unit tests
- Visual testing whilst developing

The component has lots of customisable options and elements to it.
- Banner style headings
- Headings (like titles)
- Icons (3 to pick from, Arrow, Exclamation and tick - this list can easily be extended)
- Body of text - this can accept HTML elements, so unlimited flexibility in styling and sub-components
- Input fields - although note no work has been done to implement input fields as this is not currently in any designs 

Note this is a migration/edit of the https://ons-design-system.netlify.app/components/panel/

Below is an example of some combinations of these - many more combinations exist!
panelsCombo

In addition to all this, the official colour pallet has been added to Florence in full (as it stands in dp-design-system)

### Future work

Now that this component exists the next steps is to make the first use of it on the sign-in screen in Florence, this will be used to show validation errors.

### How to review

Check the code out, maybe give the component a try.

#### But how do I use it or create a Panel?!

Glad you asked,
On a page (maybe the LoginForm.jsx) Add the following:
(Note replace {props} with actual properties you want)
```JavaScript
<Panel {props}/>
```
e.g. 
```JavaScript
<Panel type={"error"} heading={"My first Panel"} body={"wow I did it! Go me"}/>
```

Below is a list of the properties available
```JavaScript
const propTypes = {
          type: PropTypes.string.isRequired,
          body: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
          heading: PropTypes.string,
          bannerHeading: PropTypes.bool,
          icon: PropTypes.oneOf(["exclamation", "tick", "arrow"]),
          input: PropTypes.shape({
              label: PropTypes.string,
              type: PropTypes.string,
              id: PropTypes.string,
              pattern: PropTypes.string,
              inputMode: PropTypes.string,
              required: PropTypes.bool
        })
```

### Who can review

Anyone except me

✋ **New to the frontend or react or maybe just need a hand reviewing, just ask!** ✋


src/app/components/panel/__snapshots__/Panel.test.js.snap
```diff
@@ -0,0 +1,278 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
dandwal, line 1: Note this is a generated file which is used in the tests not too much reviewing to be done here.

PR #529, opened 2021-05-21
hide from user view

NoelleDL:
### What

Commented out lines that users should not be able to select. 

### How to review

Check that the following lines are hidden from user view when selecting lines in Florence:
- `ShortDashDotDot`
- `Longdashdot`
- `Longdashdotdot`

### Who can review

Anyone but me.

### DIFF
--- a/src/legacy/templates/chartEditBarlineExtras.handlebars
+++ b/src/legacy/templates/chartEditBarlineExtras.handlebars
@@ -25,9 +25,10 @@
         <option value="ShortDash">ShortDash</option>
         <option value="ShortDot">ShortDot</option>
         <option value="ShortDashDot">ShortDashDot</option>
-        <option value="ShortDashDotDot">ShortDashDotDot</option>
+        {{!-- Hidden from view since they do not display well --}}
+        {{!-- <option value="ShortDashDotDot">ShortDashDotDot</option>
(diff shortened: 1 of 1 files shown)


src/legacy/templates/chartEditBarlineExtras.handlebars
```diff
@@ -25,9 +25,10 @@
         <option value="ShortDash">ShortDash</option>
         <option value="ShortDot">ShortDot</option>
         <option value="ShortDashDot">ShortDashDot</option>
-        <option value="ShortDashDotDot">ShortDashDotDot</option>
+        {{!-- Hidden from view since they do not display well --}}
+        {{!-- <option value="ShortDashDotDot">ShortDashDotDot</option>
         <option value="LongDashDot">LongDashDot</option>
-        <option value="LongDashDotDot">LongDashDotDot</option>
+        <option value="LongDashDotDot">LongDashDotDot</option> --}}
```
jessjenkins, line 31: Just a minor formatting thing, but I think it would be clearer if individual lines were commented rather than multiple lines as a block. Ie. it's not clear initially (via git blame or similar) that `LongDashDot` is disabled unless you look at the lines above and below it.

jessjenkins (approved):

jondewijones (approved):

PR #533, opened 2021-05-27
New Validation Errors added ready to be connected to the new dp-identity-api

dandwal:
### What

Task 
 - https://trello.com/c/YzIYJHni/431-implement-validation-for-user-login

Florence currently authenticates against Zebedee. There has been work on removing this and using AWS Cognito with the new dp-identity-api sitting in-between. As part of this work, we will need to connect up the frontend when the backend is ready. It was decided that the Sign-in screen should get an update and become accessible. This work does that.
- Added new validation errors which will be thrown by the dp-identity-api
- Added usage of the new Panel component
- Refactored the signInController (which was a clone of the login controller)

Future work - Unfortunately we are using React 15 which is 4 years out of date, so the reactor was limited to what was available at the time. I have raised a new ticket in the backlog to upgrade to React 17 so we can fully use Functional React components rather than this OOP class approach.

This work has been tested with
- Keyboard navigation
- VoiceOver
- Old unit tests
- New unit tests

#### What this actually looks like
bad_unexpected_error
bad_too_many_times
bad_password
bad_email
bad_email_and_password
bad_combo

(note that due to compression and converting to a gif some colours are a little off e.g. red appears as black in some places)

### How to review

Test that the login screen behaves as expected by just running the stack like normal and having this branch checked out.

Then stop Florence and when running it again add the following to the launch command (or env variable) `ENABLE_NEW_SIGN_IN=true`.

Note although if the password and email are correct you most likely will be able to log in (as it is connected to Zebedee) errors will not be handled gracefully (this feature flag is turned off until all work on this feature including wiring it up to the dp-identity-api is complete).

Go to the src/app/views/login/SignInController.jsx
In `handleLoginError` function add errors that line up with what the dp-identity-api will send
(Information on this can be found in the swagger spec: https://github.com/ONSdigital/dp-identity-api/blob/9e5b912db12512797506a78ac9efcbd9a2be0ebb/swagger.yaml )

Typically the codes relate to the following types of validation error
```
TooManyFailedAttempts -> You've tried to sign in to your account too many times. Please try again later.
NotAuthorised -> Email address or password are incorrect
InvalidPassword -> Enter a password
InvalidField -> Enter a valid email address
``` 

E.g. Add one of the following at a time and rebuild Florence (`make node-modules; make debug ENABLE_DATASET_IMPORT=true ENCRYPTION_DISABLED=true ENABLE_NEW_SIGN_IN=true` 
```JavaScript
        error = { status: 400, errors: [{ code: "InvalidPassword" }] };
        error = { status: 400, errors: [{ code: "InvalidEmail" }] };
        error = { status: 400, errors: [{ code: "NotAuthorised" }] };
        error = { status: 400, errors: [{ code: "TooManyFailedAttempts" }] };
        error = { status: 400, errors: [{ code: "InvalidPassword" }, { code: "InvalidEmail" }] };

```
(Note the status codes above are not what they really will be but will be in the 400 range)

Validation errors will return different HTTP Status Codes between 400-499. All errors will be 500+ and success will be in the range of the 200s.

### Who can review

Anyone except me


src/app/views/login/SignInController.jsx
```diff
@@ -57,86 +58,121 @@ export class LoginController extends Component {
...
+        } else {
+            this.notifyUnexpectedError(notification);
+        }
+
+        return stateToSet;
+    }
+
+    showValidationError(anError, errorContents, notification) {
+        switch (anError.code) {
+            case "InvalidEmail":
```
mr-nick17, line 120: Does this error mean an invalid user? If so, I'd suggest renaming to `InvalidUser` and amending the text to say something like: 'Enter a valid user email address'. 

If an error is going to be returned that handles a malformed email address, keep this `case` but have another to handle the invalid user.

mr-nick17 (commented):
Possible need for an additional error case or rename existing error to be clearer

src/app/views/login/SignInController.jsx
dandwal: In the situation where a users account is invalid (password and email are in the correct format but no account is found), this will trigger 'NotAuthorised' error instead

mr-nick17 (approved):

PR #535, opened 2021-06-07
A temperary addition to the sign-in-controller

dandwal:
### What

Validation rules for dp-identity-api are between 400-499
Zebedee however uses 417 to say 'hey, you need to change your password' on login request.

Whilst feature flag is enabled, people are unable to create new users as they will be unable to login as new users need to change password on first login. In develop we need to have this feature flag turned on to get regular PO sign off as the whole of team 404 is working on auth work; so toggling the feature flag is impractacle. As other teams outside of dev team use Florence on the dev environment they are no unable to create new users on it. So this is a small solution to enable that flow of new user creation to continue. 

They will still get a toast saying "Unexpected error" and they will still get a validation error heading "Fix the following: " in the background. However users are able to change their password and progress. This isn't a long term solution. It will soon be hooked up to the dp-idtenity-api and the 'change password' journeys are not yet created.

Needed as dev environment is actively in use and new users need to be added but we are mid process of changing the backend.

Note this has no effect on production as this screen is only hit if feature flag is enabled (`ENABLE_NEW_SIGN_IN=true`) production still uses the old login controller and so the user creation journey is 100% working as expected currently and this PR will have no effect on production - as the feature flag is disabled in prod until all work is complete.

### Who can review

Anyone except me

### DIFF
--- a/src/app/views/login/SignInController.jsx
+++ b/src/app/views/login/SignInController.jsx
@@ -105,6 +105,11 @@ export class LoginController extends Component {
                 };
                 stateToSet.email.errorMsg = errorContents.emailMessage;
                 stateToSet.password.errorMsg = errorContents.passwordMessage;
+                // Temporary code until login work is connected to dp-identity-api
+                if (error.status === 417) {
+                    stateToSet.requestPasswordChange = true;
+                }
+                // End of temporary code
             } else {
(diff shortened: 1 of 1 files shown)


cadmiumcat (approved):
This looks good 👍

PR #538, opened 2021-06-09
Hook up Sign Out

dandwal:
### What

Although the dp-api-router needs to be updated for this to work
- Fix bug where on cookie delete on sign out - if two requests 
- Fix bug that notifications would be deleted if user was signed out as it would set the state to initial state, which means notification array was set to empty immediately after adding a sign out notification error
- Http-requester now allows headers to be set

Bug still present but not part of this PR but now visible due to fixing other bugs (note not a functional issue just visual annoyance) if Florence sends two API requests at once and the users auth token has been removed then these requests will both return errored (401) however they both will sign the user out so you get double notifications saying this has happened - raising a card in the backlog around this.

### How to review

Run Florence without the `ENABLE_NEW_SIGN_IN` feature flag. Ensure that you are able to sign out of Florence like normal.

Run Florence with the feature flag above enabled. Note that Sign out will error but you will be signed out. Let me explain.
Florence used to sign you out by just deleting the session cookie client side, so it's like losing your keys to a lock.
Now Florence will actually tell the owner of the lock 'hey, I no longer need my keys, I am going to dispose of them'. The lock owner is not very trusting of Florence says 'ok' but then changes the lock too so that the keys Florence has will no longer work. Then Florence goes 'cool now time to loose my keys' and does so. 

But note there could be a server error when telling the lock owner Florence doesn't need the keys anymore when this happens an error will trigger BUT Florence will continue to 'lose its keys'. The user at this point will be directed back to the sign in screen and a toast will notify them an error occurred during sign out please contact a system administrator - at this point we should investigate why this error happened.

Right now this error will always happen as requests will go to the dp-api-router but the router isn't configured yet proxy requests to the dp-identity-api (there is a task in the backlog to hook that up - but depends on another task, versioning the endpoints).

### Who can review

Anyone except me

### DIFF
--- a/src/app/utilities/api-clients/user.js
+++ b/src/app/utilities/api-clients/user.js
@@ -1,7 +1,9 @@
 import http from "../http";
 import { store } from "../../config/store";
-import { userLoggedIn, userLoggedOut, reset } from "../../config/actions";
+import { reset, userLoggedIn, userLoggedOut } from "../../config/actions";
 import cookies from "../cookies";
+import notifications from "../notifications";
+import log from "../logging/log";
 
 export default class user {
     static get(email) {
@@ -30,6 +32,10 @@ export default class user {
 
     static getPermissions(email) {
         return http.get(`/zebedee/permission?email=${email}`, true, true);
+    }
+
+    static expireSession(headers) {
+        return http.delete("/tokens/self", true, true, headers);
     }
 
     static getOldUserType(user) {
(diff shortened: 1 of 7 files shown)


src/app/utilities/api-clients/user.js
```diff
@@ -72,18 +78,56 @@ export default class user {
...
+                    if (error.status === 400) {
+                        const notification = {
+                            type: "warning",
+                            message: "An error occurred during sign out 'InvalidToken', please contact a system administrator",
+                            isDismissable: true,
+                            autoDismiss: 20000
+                        };
+                        notifications.add(notification);
+                        console.error("Error occurred sending DELETE to /tokens/self - InvalidToken");
+                        log.event("error on sign out sending delete to /tokens/self failed with an invalid token", log.error(error));
```
rav-pradhan, line 114: Not something for this PR but looking at this, it would be good to update the logging package we use in Florence to align with our logging standards with the `log.go` v2 changes. We could also then maybe introduce new functions to also append warnings/errors etc. to the console rather than having to make two separate calls like here

rav-pradhan (approved):
Tested locally and I could see the behaviour as described in the PR. Looks good

src/app/utilities/api-clients/user.js
dandwal: Agreed, raised a ticket in the backlog for this: https://trello.com/c/kHlQt53r/2110-florence-js-logging-needs-to-be-updated-to-be-in-line-with-current-logging-standards

PR #539, opened 2021-06-14
npm audit fix run in both src/ and src/app

dandwal:
### What

Audit was failing, so ran npm audit fix in both src/ and src/app to update dependencies and hopefully resolve security issues

### Who can review

Anyone except me


rav-pradhan (approved):
I ran the audit locally and it passed, assuming it will be the same once the job in Concourse finishes 😅

PR #545, opened 2021-06-30
Wire up new sign in functionality

dandwal:
### What

- New routing path added to server-side Florence /tokens
- id, refresh and access token now set on sign in into a cookie
- removed auth header being added client-side on outbound reqeusts, this happens on server side Florence so is not needed here
- Removed unused 'headers' property for fetch wrapper, since refactoring of above point this is unused so has been removed
- error that was never logged out is now logged out if server-side Florence is unable to set auth token header
- Cookie util can now set additional properties, Secure and SameSite
- go version incremented
- npm audit fix run in src/ and src/legacy

Note when the feature flag is enabled `ENABLE_NEW_SIGN_IN` a request is made to dp-identity-api to sign in rather than to Zebedee. Straight after this request, Florence attempts to access the /permissions endpoint of Zebedee however this request will fail. As such from a users point of view, you will see that there was a failure to sign in and a toast will appear explaining this. this is as expected. If we look at the network requests we can see that a request to 201 will succeed if you use a valid login for dp-identity-api, please ask if you require one - you can't use your usual Florence login as this is only registered against Zebedee.

Once the request is made we store the three returned tokens in cookies:
`refresh_token`
`access_token`
`id_token`

At the moment Florence is incredibly insecure. These three cookies are stored in the same manner the Zebedee token was stored as part of this task. There is another task to make these secure coming up, as part of that task access_token and refresh_token will be set via server-side Florence rather than client-side. As this is a public repository I won't divulge too much information on here around this but please feel free to ask if you need more information on this matter.

In previous work, we were setting the header on the client-side for authentication. This was actually incorrect and was unneeded but only discovered during this ticket. server-side Florence will set the header. As such this, newer but uneccesarry functionality has been ripped out.

When the feature is enabled we can see it working in the developer tools network tab client side:
Screenshot 2021-06-30 at 12 04 08

We can also see the cookies are set by going to `tokens/self` (note this is not a visual endpoint but as refresh cookie path is set to just /tokens/self as we don't want it to be exposed except when required you can go to the endpoint and see it. Otherwise, in Florence you will just see the id token and access token as their path is set to `/`:
Screenshot 2021-06-30 at 12 15 42

At the moment in this PR the cookies are set client-side, as was the way prior to this PR. This is bad. Auth cookies or any cookies with sensitive information should only ever be set server-side (via the `set-cookie` header which tells the browser to set the cookie). This is because the httpOnly attribute should be set. When this is set it will prevent JS from ever being able to access the cookie, which helps prevent XSS attacks and mitigate against CSRF attacks. It was decided that this was going to be split out into a second PR as it involves updating a number of our libraries to enable us to set these in server-side Florence. 

### How to review

- Check that when the feature flag is disabled that you can log into Florence like normal
- Check the code looks ok

### Who can review

Anyone except me


src/app/utilities/cookies.js
```diff
@@ -10,7 +10,7 @@ export default class cookies {
         return cookiesObject;
     }
 
-    static add(name, value, path, domain) {
+    static add(name, value, path, domain, sameSite, secure) {
```
mr-nick17, line 13: This is the upper limit of parameters for a method signature, consider breaking into a object

mr-nick17, line 13: This method has got 6 `if` statements. Consider breaking out the `error` handling into a separate function as these statements are not "`add`ing" anything (as per method name).

mr-nick17 (commented):
5 parameters are the upper limit for method signatures, appreciate some of it is existing code but the `add` method is ripe for refactoring.

No unit tests for `cookies.js`?

dandwal:
> 5 parameters are the upper limit for method signatures, appreciate some of it is existing code but the `add` method is ripe for refactoring.
> 
> No unit tests for `cookies.js`?

I have added tests for the add function in the spirit of 'clean campsite', whoever touches the other functions should create tests for those. 

Although note in the following PR the Auth cookies will no longer be set client-side (they never should have to be honest as they are vulnerable, this PR sets them as we decided to secure them we will make a new ticket in sprint as it requires reworking the reverseProxy and a few other libraries of ours so for the time being (in this PR) we will continue to set them client side)

mr-nick17 (approved):

PR #546, opened 2021-07-12
Forgot password link added to Sign in screen

dandwal:
### What

Florence's new Sign-in screen has a new link "Forgotten your password" which takes the user to `/florence/forgotten-password`. The page it takes you to, does not yet exist, but, will be created in future work.

The 404 Florence page used to say "Sorry, this page couldn't be found" the 'couldnt' was bugging me so added an apostrophe so it now read "couldn't"

Took the branch from master rather than develop, but there are no code changes that are different

### How to review

Run Florence, there should be no link on the screen for "Forgotten your password?" on the sign-in screen.

Run Florence again but this time add the feature flag `ENABLE_NEW_SIGN_IN=true` then go to the sign-in screen. There should be a link above the Sign-in button saying "Forgotten your password".

Test tabbing, there should be a yellow box around it when the link is the focus element. Try clicking space/enter (depends on your local mac keyboard bindings) once focused it should take you to the page `/florence/forgotten-password`. Although note the page will just say "Sorry, this page couldn't be found.

### Who can review

Anyone except me

### DIFF
--- a/src/app/views/login/SignInForm.jsx
+++ b/src/app/views/login/SignInForm.jsx
@@ -36,6 +36,9 @@ const LoginForm = props => {
                     {props.inputs.map((input, index) => {
                         return <Input key={index} {...input} disabled={props.isSubmitting} allowAutoComplete={true} />;
                     })}
+                    <div>
+                        <a href={"/florence/forgotten-password"}>Forgotten your password?</a>
+                    </div>
                     <button type="submit" className="btn btn--primary margin-top--1" disabled={props.isSubmitting}>
                         Sign in
                     </button>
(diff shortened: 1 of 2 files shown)


cadmiumcat (commented):
I had a go:
- In Chrome this works as you described
- In Firefox I can't focus on "Forgotten your password?" (there's no yellow around it). I have to click on it, rather than press enter, for it to work

Neither of them shows a yellow box around the "show password" element, I'm not sure if that's intentional

cadmiumcat (approved):
After discussing - Users will only use Chrome anyway

PR #547, opened 2021-07-15
Secure Auth cookies

dandwal:
#### Description

As part of the Auth work, we have set the new cookies for the dp-identity-api.
Whilst doing this work it was discovered that our cookies are not at all secure.
As such this PR work spawned from this. Cookies with sensitive info should be set server-side and not by the client; As they must have httpOnly set.

#### New, Changed or Removed items

- New modifyResponse files and tests added, these are used by the reverseProxy
- Updated the dp-net package to v2 which now allows our reverseProxy wrapper to accept a modifyResponse function, this is called when a response from an outbound request is received and before returning it to the requester.
- modifyResponse function added that takes the Auth headers if present and adds them to 'set-cookie'
- Removed Client setting Auth cookies
- Sever-side Florence now sets the Auth cookies in a secure and correct manner
- Updated the dp-api-clients-go to v2 
- Updated files to be formatted correctly

Below we can see cookies set correctly with the correct attributes
Screenshot 2021-07-08 at 10 37 40

### How to review

- Check out this branch
- Run Florence without the feature flag `ENABLE_NEW_SIGN_IN` on and ensure you can sign in and use Florence like normal.
- Run Florence with the feature flag `ENABLE_NEW_SIGN_IN=true` and try to sign in. Note that to sign in with this feature flag enabled you will need to use an account associated with AWS Cognito, not your typical Zebedee account. This auth work is to remove Zebedees Auth handling and move it all to AWS Cognito.
Note: To see the refresh token in your network tab you may have to go to the URL `tokens/self` as the Refresh token is only sent on that path for better security (I believe some browsers will only show in the dev tools the cookie if you are on the same domain, others only if on same path and domain - I may be wrong though)

### Who can review

Anyone except me


service/modifier.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"net/http"
+)
+
+// Edits the response before proxying it back.
+// If Auth tokens are present then set them as cookies
+func identityResponseModifier(r *http.Response) error {
+	// maxAgeBrowserSession is length of time to expire cookie on browser close
+	maxAgeBrowserSession := 0
+	if r.StatusCode >= 200 && r.StatusCode < 300 {
```
mike-kelly-methods, line 12: Since your're importing the `net/http` package, its probably worth using the the `http` constants rather than inlining the codes:

```
if r.StatusCode >= http. StatusOK && r.StatusCode < http. StatusMultipleChoices {
```

dandwal: Good shout

mike-kelly-methods (approved):
👍

PR #548, opened 2021-07-15
Workaround for race issue with generating main.min.css

jessjenkins:
### What

The node step of the build was completing before the asset was generated meaning the next step to generate go assets wasn't including it.
Introduced a sleep to allow time for the asset to complete.

### How to review

Ensure it builds OK on develop.

### Who can review

Anyone but me


mr-nick17 (approved):
LGTM - raise ticket to investigate why npm install is reporting completion when .css file is still being written

PR #555, opened 2021-07-30
Feature/change password

sandypadmanabhan:
### What
https://trello.com/c/2k5u5rxe/511-your-password-has-been-changed-screen-first-login
Added a new page as per ticket

### How to review
run the publishing stack  
run florence with ENABLE_NEW_SIGN_IN=true
 update the state.hasSubmitted to true before you run
 /src/app/views/change-password/changePasswordController.jsx - line 14

go to http://localhost:8081/florence/change-password and check if the screen looks as per ticket

### Who can review

anyone


src/app/views/change-password/changePasswordConfirmed.jsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Panel from "../../components/panel/Panel";
+
+const propTypes = {
```
dandwal, line 5: This page doesn't actually take any properties so we can remove all the propTypes

src/app/views/change-password/changePasswordConfirmed.jsx
```diff
@@ -0,0 +1,41 @@
...
+                <h1> Your password has been changed</h1>
+                <a href="/florence/collections" className="margin-bottom--2 margin-right--5 anchor-btn">
+                    Start using Florence
+                </a>
+                <Panel type={"information"} body={changePasswordPanelBody} />
+            </div>
+        </div>
+    );
+};
+ChangePasswordConfirmed.propTypes = propTypes;
```
dandwal, line 40: As this page has no properties we don't need to include this

src/app/views/change-password/changePasswordConfirmed.jsx
```diff
@@ -0,0 +1,41 @@
...
+        type: PropTypes.string,
+        onChange: PropTypes.func,
+        error: PropTypes.string
+    })
+};
+
+const ChangePasswordConfirmed = props => {
+    const changePasswordPanelBody = (
+        <p>
+            If you have technical problems with your account, please email
```
dandwal, line 23: I can't quite tell from looking at the code (will need to run it to test) but think there is a missing space before the anchor tag. This is needed or we will see:
`If you have technical problems with your account, please emailpublishing@ons.gov.uk`
rather than 
`If you have technical problems with your account, please email publishing@ons.gov.uk`

src/app/views/change-password/changePasswordConfirmed.jsx
```diff
@@ -0,0 +1,41 @@
...
+            <a href="mailto:publishing@ons.gov.uk">publishing@ons.gov.uk</a>
+        </p>
+    );
+
+    return (
+        <div className="grid grid--justify-center">
+            <div className="grid__col-3">
+                <h1> Your password has been changed</h1>
```
dandwal, line 31:
```suggestion
                <h1>Your password has been changed</h1>
```

src/app/views/change-password/changePasswordRequest.jsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
```
dandwal, line 1: Note, haven't reviewed this file other than it compiles as this is to be constructed in a future task

src/app/views/change-password/changePasswordRequest.jsx
```diff
@@ -0,0 +1,30 @@
...
+        error: PropTypes.string
+    })
+};
+
+const ChangePasswordRequest = props => {
+    // TODO this page is to be done in an upcoming task
+    return (
+        <div className="grid grid--justify-center">
+            <div className="grid__col-3">
+                <h1>We are on change password request</h1>
```
dandwal, line 24: As this will be deployed to prod (although behind a feature flag) we need to make sure on the off chance someone stumbles upon the page before we change this heading that they don't see irrelevant comments. We should probably change this h1 to say something like "Request password change" as "We are on change password request" is quite informal.

src/scss/elements/_anchor.scss
```diff
@@ -0,0 +1,11 @@
+.anchor-btn{
+    background-color: #337099;
+    color: white !important;
```
dandwal, line 3: Do we need the `!important` line here?

Typically (not always, there are some rare instances where it is fine) `!important` should be avoided as a rule of thumb

src/scss/elements/_anchor.scss
```diff
@@ -0,0 +1,11 @@
+.anchor-btn{
+    background-color: #337099;
```
dandwal, line 2: We should use a variable here which we already have for this colour in the _variables.scss file
```suggestion
    background-color: $astral;
```

dandwal (commented):
I don't see mention of Focus, and Hover styling for this button do we maybe need to add these? 

Also, have we tested this page with keyboard navigation ('Tab' to go to the next element, 'Shift+Tab' to go back to an element? In addition, might be worth checking 'VoiceOver' on the page (under acessability settings on the Mac)

src/scss/elements/_anchor.scss
sandypadmanabhan: unfortunately, if I remove important, the font colour is black

src/app/views/change-password/changePasswordRequest.jsx
sandypadmanabhan: not sure about this comment. Ta

src/scss/elements/_anchor.scss
dandwal: So the black colour is being applied by the pseudo class `visited`
```scss
a:visited {
    color: $pickled-bluewood;
}
```

So all generic `a` tags that have the visited state applied will apply the pseudo class `:visited`

As this is a button, we don't want that pseudo class styling to be applied, so we can add something like:

```scss
.anchor-btn {
    &:visited {
        color: white
    }
    background-color: #337099;
    color: white;
    display: inline-block;
    border-radius: 4px;
    padding: 15px 25px;
    font-size: 18px;
    text-decoration: none;
    max-width: 215px;
    min-width: 100px;
  }
```

This works as a class selector  (anything that starts with a `.` is more specific than a generic type selector such as just `a`

The `&` here just means copy parent name so as the parent of `&:visited` is `.anchor-btn` the `&` will be replaced when transpiled from SCSS to CSS as `.anchor-btn:visted`

src/app/views/change-password/changePasswordConfirmed.jsx
```diff
@@ -0,0 +1,26 @@
...
+            If you have technical problems with your account, please email
+            <a href="mailto:publishing@ons.gov.uk"> publishing@ons.gov.uk</a>
+        </p>
+    );
+
+    return (
+        <div className="grid grid--justify-center">
+            <div className="grid__col-3">
+                <h1>Your password has been changed</h1>
+                <a href="/florence/collections" className="margin-bottom--2 margin-right--5 anchor-btn">
```
dandwal, line 17: Font size should be 14px here according to the designs

src/scss/elements/_anchor.scss
dandwal: Although note #337099 should also be a variable `$ash` I believe matches that colour

dandwal (dismissed):
We have gone through it and can see, keyboard navigation is working, VoiceOver is working and the code looks fine.

dandwal (approved):

PR #568, opened 2021-08-13
Release 1.37.0

dandwal:
### What

Release 1.37.0

- Forgotten password screen complete
- Tidy up
  - lets to const
  - Change numerous JSX  `={"x"}` to `="x"` for performance gains
  - Fixed spelling of `heading`
  - Fixed misuse of Panel component

### Who can review

Anyone except me

### DIFF
--- a/src/app/views/forgotten-password/forgottenPasswordController.jsx
+++ b/src/app/views/forgotten-password/forgottenPasswordController.jsx
@@ -1,6 +1,4 @@
 import React, { Component } from "react";
-import { connect } from "react-redux";
-import { push } from "react-router-redux";
 import PropTypes from "prop-types";
 import ForgottenPasswordRequest from "./forgottenPasswordRequest";
 import ForgottenPasswordEmailSent from "./forgottenPasswordEmailSent";
@@ -9,10 +7,7 @@ import { errCodes } from "../../utilities/errorCodes";
 import notifications from "../../utilities/notifications";
 
 const propTypes = {
-    dispatch: PropTypes.func.isRequired,
-    isAuthenticated: PropTypes.bool.isRequired,
-    rootPath: PropTypes.string.isRequired,
-    location: PropTypes.object
+    dispatch: PropTypes.func.isRequired
 };
 
 export class ForgottenPasswordController extends Component {
@@ -20,10 +15,6 @@ export class ForgottenPasswordController extends Component {
         super(props);
         this.handleInputChange = this.handleInputChange.bind(this);
         this.handleSubmit = this.handleSubmit.bind(this);
-        this.requestEmailChange = this.requestEmailChange.bind(this);
-        this.postResetPassword = this.postResetPassword.bind(this);
-        this.handleResetPasswordError = this.handleResetPasswordError.bind(this);
-        this.showValidationError = this.showValidationError.bind(this);
         this.state = {
             validationErrors: {},
             email: {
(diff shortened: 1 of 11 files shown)


src/app/views/forgotten-password/forgottenPasswordController.test.js
```diff
@@ -0,0 +1,105 @@
...
+    });
+    it("Show the users email", () => {
+        expect(
+            component
+                .find("strong")
+                .at(0)
+                .text()
+        ).toBe("foo@bar.com");
+    });
+});
```
bpathak-ons, line 105: I think tests also include fail scenarios for email and password

src/app/views/forgotten-password/forgottenPasswordController.jsx
```diff
@@ -75,13 +79,15 @@ export class ForgottenPasswordController extends Component {
...
                     error.body.errors.forEach(anError => {
                         errorContents = this.showValidationError(anError, errorContents);
-                        validationErrors.heading = "Fix the following: ";
                     });
                 } else {
                     this.notifyUnexpectedError();
                 }
 
-                validationErrors.body = errorContents.errorsForBody;
+                if (errorContents.errorsForBody.length >= 1) {
```
bpathak-ons, line 87: we should capture the error with email/userId for logging purpose

src/app/views/login/SignInController.jsx
```diff
@@ -133,19 +133,21 @@ export class LoginController extends Component {
...
                 );
                 errorContents.passwordMessage = "Enter a password";
                 break;
             case "NotAuthorised":
-                errorContents.errorsForBody.push(<p>Email address or password are incorrect</p>);
+                errorContents.errorsForBody.push(<p key="error-not-authorised">Email address or password are incorrect</p>);
                 break;
             case "TooManyFailedAttempts":
-                errorContents.errorsForBody.push(<p>You've tried to sign in to your account too many times. Please try again later.</p>);
+                errorContents.errorsForBody.push(
```
bpathak-ons, line 148: if possible, pls look at my previous comment regarding "capture the error with email/userId"

src/app/views/forgotten-password/forgottenPasswordController.test.js
dandwal: Thanks for your review.

If the email is invalid then we don't tell the user - as this would be a security risk so the scenario is treated as a success - user is presented with a second screen saying if their email was valid then they will have an email, if they didn't get one but expected one then they need to contact an admin.

The backend can return a validation error which is handled in the test `When the user submits their email for a password request but the server returns a validation error`

The HTTP wrapper we use wraps around another wrapper (Request) that handles and logs errors, it also attempts numerous retries)

These tests are testing the react 'state' marry up to the visual elements being displayed. We have a task coming in the next sprint to add feature tests to Florence which will check each journey and scenario in more length. These feature tests will be added once the whole journey has been complete, at the moment this is only part of the journey and is disabled behind a feature flag so will not be reachable in production when released.

src/app/views/login/SignInController.jsx
dandwal: This release doesn't really change the functionality of the SignInController, but I made a few cleanup changes only.
However, the Invalid Email/UserID is handled in the NotAuthorised case. We do not specify which is incorrect as this would be a security risk.

src/app/views/forgotten-password/forgottenPasswordController.jsx
dandwal: The error is captured on Florence Serverside (within this repo, it will output to our centralised logging system which can then be accessed in Kibana). Once server-side error has been logged it then passes sends the clientside a response which is what we see here so that the user can see something went wrong. In the case of a bad email/password these are handled in the showValidationError switch statements.

bpathak-ons (approved):
looks ok to me

PR #569, opened 2021-08-19
Refresh and ID cookie values injected as headers on refresh requests

dandwal:
### What

This PR is part of the wider new Auth work to replace Zebedee authentication with AWS Cognito via the dp-identity-api

Secure cookies that store ID and Refresh Tokens must be appended to the outbound requests leaving Florence on the path /tokens/self. These requests go to the dp-identity-api and once recieved will return a new session token (aka access token) once that session token has been provided this will replace the user's current session token and be used for all API requests.

### How to review

Check tests pass.
If you want to test the functionality completely then there are a few steps to jump through, but I would probably say this isn't needed for the PR.

 In Florence run `make node-modules` and then proceed to start Florence server-side with the feature flag `ENABLE_NEW_SIGN_IN=true`

Then create a new account with a REST client (e.g. ARC or POSTMAN)

```
Endpoint: http://localhost:25600/v1/users
Method: POST
Body: {
  "forename": "something",
  "lastname": "somethingElse",
  "email": "someRandomEmail@ons.gov.uk"
}
```

Note please replace `forename` `lastname` and email values in the body with whatever you choose but the email must end with `@ons.gov.uk`
Run the following command in terminal (note require aws CLI installed): 'aws cognito-idp admin-set-user-password --user-pool-id [pool-ID] --username [user-id]  --password [desired-new-password] --permanent'

Where `[pool-id]` matches the user pool of the dp-identity-api (on startup this is outputted, or look at the code configs for this value)
`[user-id]` is the user id outputted from the HTTP request we sent in the previous step
`[desired-new-password]` is a new password you want. It must be more than 14 characters, have at least 1 upper case, 1 lowercase and 1 number. Or the request will be rejected (can read the output it will tell you something is wrong).

Once complete you can load up Florence login screen locally e.g. `http://localhost:8081/florence/login` and open the network tab in the dev tools.

Proceed to enter the email and password that you have set in the above steps. You will notice an error pop up at the bottom of the screen when you click `Sign in`.

When the button is pressed two requests are made, a post sign-in request and a 
get permissions request. The latter one should be the one that fails, you can identify this in the network tab. So long as the first one was a success you will be given three tokens `ID`, `Refresh` and `Session`  these are stored in your cookies, but can't be accessed via JS as they are secure httpOnly cookies. As such only Server-Side applications can access them, which is where this PR comes in.

Finally, in the dev tools, you can open a console and enter the following code which will make a request to refresh the current user's session token by providing the current ID and Refresh Tokens
```Javascript
const fetchConfig = {
    method: "PUT",
    body: JSON.stringify({})
  };
fetch("/tokens/self", fetchConfig)
  .then(response => {
  	console.log('Well got a response')
	  if (!response.ok) {
    	throw response;
  	}
  	console.log('response was ok')
  	return response.json();
})
```

If you tab over to the window you are running the dp-identity-api you should notice it returning a `status 201`

### Who can review

Anyone except me


rav-pradhan:
Looks good, but the audit check is failing due to the `copy-webpack-plugin` - may be worth bumping the version and see if that resolves

rav-pradhan (approved):

PR #579, opened 2021-09-08
New forgotten password process all linked up

dandwal:
### What

This PR is part of the wider new Auth work to replace Zebedee authentication with AWS Cognito via the dp-identity-api

This is the final screen in the "Forgotten password" user journey.
This also replaces a few areas that used separate button and spinner components and now uses a new combined one

Preview (note poor quality due to compression)
passwordPreview

### How to review

Check tests pass.
(There is a hardcoded UID at the moment in the email templates, when it comes to reviewing this PR please let me know so I can change this in AWS (this part of the work is WIP by other tickets).

To check fully there are a few steps that need to be jumped through, but these guided steps should give you all that you need.

Start up the dp-identity-api by running: `make debug ALLOWED_EMAIL_DOMAINS="@ons.gov.uk,@ext.ons.gov.uk,@gmail.com"` Please note that you will need to use an email for a latter step that you have access to the inbox of, please replace the `ALLOWED_EMAIL_DOMAINS` with an email domain you will later use for this.

In Florence run `make node-modules` and then proceed to start Florence server-side with the feature flag `ENABLE_NEW_SIGN_IN=true`

Then create a new account with a REST client (e.g. ARC or POSTMAN)

```
Endpoint: http://localhost:25600/v1/users
Method: POST
Body: {
  "forename": "something",
  "lastname": "somethingElse",
  "email": "youremail@gmail.com"
}
```
(Note you must be able to access whatever email inbox you are using for this test and you will need to ensure the domain is valid, matching one you used to start the application with in the first step).

Run the following command in terminal (note require AWS CLI installed): 'aws cognito-idp admin-set-user-password --user-pool-id [pool-ID] --username [user-id]  --password [desired-new-password] --permanent'

Where `[pool-id]` matches the user pool of the dp-identity-api (on startup this is outputted, or look at the code configs for this value)
`[user-id]` is the user id outputted from the HTTP request we sent in the previous step
`[desired-new-password]` is a new password you want. It must be more than 14 characters, have at least 1 upper case, 1 lowercase and 1 number. Or the request will be rejected (can read the output it will tell you something is wrong).

Once complete you can load up Florence login screen locally e.g. `http://localhost:8081/florence/login`

Then click forgotten password
Then enter your email address
Go to your inbox
Click the link (note it will likely take you to develop publishing (this is fine, the email template is a bit WIP in another task and will later be using a lambda) copy everything in the URL from `/florence` onwards and then go back to `http://localhost:8081/florence/` and replace `/florence` with the contents of your clipboard. The URL now looks something like: `https://publishing.develop.onsdigital.co.uk/florence/password-reset?vid=123456&uid=1234a56b-f0dc-478d-b065-9bfdac0000f0`

Test this screen, as this is the screen made in this PR.

### Who can review

Anyone except me


src/app/views/forgotten-password/setForgottenPasswordRequest.jsx
```diff
@@ -1,14 +1,30 @@
...
 
 const SetForgottenPasswordRequest = props => {
-    // TODO this page is to be done in an upcoming task
     return (
         <div className="grid grid--justify-center">
             <div className="grid__col-3">
-                <h1>Request password set</h1>
+                <h1>{props.heading}</h1>
+                <form className="form" onSubmit={props.onSubmit}>
+                    <ValidateNewPassword updateValidity={props.validityCheck} inputError={props.showInputError ? " " : ""} />
```
dandwal, line 21: Aware this is a little hacky, inputError is passed down to an Input component which if it has inputError it will make the input red (errored state) but as per the UX designs, we don't want the input error text to show.

src/app/views/forgotten-password/setForgottenPasswordController.jsx
```diff
@@ -11,12 +15,115 @@ export class SetForgottenPasswordController extends Component {
     constructor(props) {
         super(props);
         this.state = {
-            hasSubmitted: false
+            hasSubmitted: false,
```
AnnaEmson, line 18: I don't think you need all that on state, you can make constant outside the class so it is not created on every render. 
`const STATUS = { SUBMITTED : 'SUBMITTED',  SUBMITTING: 'SUBMITTING', .... etc } `
   and then when you modify just reference `this.setState({ STATUS.SUBMITTED });` and state would only hold one status at the same time , you also probably need completed etc .. and idle?

src/app/components/validation-item-list/ValidationItem.jsx
```diff
@@ -26,8 +26,9 @@ export default class ValidationItem extends Component {
                     onClick={this.preventUserInteractions}
                     onKeyDown={this.preventUserInteractions}
                     className="readonly-checkbox"
+                    tabIndex={-1}
```
AnnaEmson, line 29: just curious  why are you removing this element from the default navigation flow, users will not be able to tab on it using the keyboard?

src/app/components/validation-item-list/ValidationItem.jsx
```diff
@@ -26,8 +26,9 @@ export default class ValidationItem extends Component {
                     onClick={this.preventUserInteractions}
```
AnnaEmson, line 26: I think you should have onChange even here

src/app/views/forgotten-password/setForgottenPasswordRequest.jsx
AnnaEmson: I would investing the possibilities to remove this hack. can you maybe pass a boolean instead?

src/app/components/validate-new-password/ValidateNewPassword.jsx
```diff
@@ -155,6 +156,7 @@ export class ValidateNewPassword extends Component {
                     onChange={this.handleInputChange}
                     disableShowPasswordText={true}
```
AnnaEmson, line 157: I think you do not need so complex state. the type of this input is always the same. can you please move it form state to `type='password' ` here. I think this will help simplify the state. It is important to keep the minimum possible in state.

src/app/components/validate-new-password/ValidateNewPassword.jsx
```diff
@@ -7,7 +7,8 @@ import notifications from "../../utilities/notifications";
 import ValidationItemList from "../../components/validation-item-list/ValidationItemList";
 
 const propTypes = {
-    updateValidity: PropTypes.func
+    updateValidity: PropTypes.func,
+    inputError: PropTypes.string
 };
 
 export class ValidateNewPassword extends Component {
```
AnnaEmson, line 14: please simplify the state here. no need for complex object here for password. keep only dynamic values in the stae, remove the validation rules as they do not change.

src/app/views/forgotten-password/setForgottenPasswordController.jsx
```diff
@@ -11,12 +15,115 @@ export class SetForgottenPasswordController extends Component {
...
+            validityCheck: this.validityCheck,
+            isValid: this.state.passwordIsValid,
+            onSubmit: this.onSubmit,
+            heading: "Create a new password",
+            buttonText: "Confirm password",
+            showInputError: this.state.showInputError,
+            isSubmitting: this.state.isSubmitting
+        };
+
+        const screenToShow = this.state.hasSubmitted ? (
```
AnnaEmson, line 122: i would probably simplify this a bit:
if you make changes as per my comment above: 
`if( this.state.status === STATUS. SUBMITTED) return  <SetForgottenPasswordConfirmed /> 
return <SetForgottenPasswordRequest {...setForgottenPasswordRequestProps} />` don't store this in constant, no need for it :-)

src/app/views/forgotten-password/setForgottenPasswordController.jsx
```diff
@@ -11,12 +15,115 @@ export class SetForgottenPasswordController extends Component {
...
+        if (!this.state.passwordIsValid) {
+            this.setState({ showInputError: true });
+            return;
+        }
+        let verificationID = new URLSearchParams(location.search).get("vid");
+        let userID = new URLSearchParams(location.search).get("uid");
+        // UID is sent in Email field, for speed of first delivery it was decided UID is to be used instead of email.
+        // Eventually the backend should be changed to use the endpoint /users/{uid}/password or at least change the
+        // email field to be uid and then (TODO) this can be updated
+        const requestBody = {
```
AnnaEmson, line 39: can you maybe move this out of here I think it is part of the request not submit

src/app/views/forgotten-password/setForgottenPasswordController.jsx
```diff
@@ -11,12 +15,115 @@ export class SetForgottenPasswordController extends Component {
...
+        notifications.add(notification);
+    }
+
+    validityCheck(isValid, password) {
+        // Only show input error if user had previously tried to submit password and it is still invalid
+        const showInputError = !isValid && this.state.showInputError;
+        this.setState({
+            passwordIsValid: isValid,
+            password: password,
+            showInputError: showInputError
```
AnnaEmson, line 107: I think this can be simplify too. you don't need two properties `passwordIsValid ` and separate `showInputError` they represent the same situation. If the password is not valid show the error. you don't show the error when password is valid. state must be as simple as possible and only contain elements which change need trigger re render.

src/app/views/forgotten-password/setForgottenPasswordController.jsx
dandwal: We have been using `hasSubmitted` and `isSubmitting` in other places, but yes I think you are right it would be better to use an enumerator to store the state as it will never be both `hasSubmitted` and `isSubmitted` at the same time.

src/app/components/validation-item-list/ValidationItem.jsx
dandwal: This component is a read-only type of checkbox, so the user won't be able to interact with it, just read its state which is set in code based on their password input. As  `onKeyDown={this.preventUserInteractions}` is used it prevents keyboard users from using the space key to edit the state of the checkbox. As a side-effect, this also creates a keyboard trap. As it is non-interactable I change the tab index to -1 as the user should not be able to edit it. However, thinking a bit more perhaps we should just block space key code that way the user should be able to use voice-over technology to still read the checkboxes state. If they choose to.

We have some work upcoming to update the notification bar (on error this appears) I am planning on inserting an aria alert on the error type.  Which would inform the audio only users of the validation error on this screen too. So they aren't reliant on these verification checkboxes.

src/app/components/validation-item-list/ValidationItem.jsx
dandwal: This is a read-only checkbox so the user is not able to interact with it and so the onChange function will never be fired. :)

src/app/components/validate-new-password/ValidateNewPassword.jsx
dandwal: The state does change based on if the user clicks show password, By default is of type password so that the password remains hidden. When the user clicks show password, the type changes to text

src/app/components/validation-item-list/ValidationItem.jsx
dandwal: Cool so, doing something like the following seems to do the trick (where e is the event):
```JavaScript
    const spaceKeyCode = 32;
        if (e.keyCode === spaceKeyCode || e.type == "click") {
            e.preventDefault();
        }
```

AnnaEmson: maybe use just `disable` attribute?

src/app/components/validate-new-password/ValidateNewPassword.jsx
AnnaEmson: oh, that make sense, but maybe flatten the state

src/app/components/validation-item-list/ValidationItem.jsx
AnnaEmson: so what is onClick doing there? :-)

src/app/components/validate-new-password/ValidateNewPassword.jsx
AnnaEmson: Also I would probably think about changing the name of the component NewPasswordInput.jsx.or just so in that case having state={type: 'password', password: '' .... }  would make sense .. or something similar. you can create seperate function to take password and pass it thru validation rules, if they don't pass show an error.

src/app/components/validate-new-password/ValidateNewPassword.jsx
AnnaEmson: can you please change the event on checkbox form handleInputChange to TogglePasswordVisibility and only change the type in state , nothing else. keep it simple. :-) Functions needs to be pure and do only one thing at the time :-)

src/app/components/validation-item-list/ValidationItem.jsx
AnnaEmson: so as above `readonly="readonly" `should do the job :-)

src/app/views/forgotten-password/setForgottenPasswordRequest.jsx
dandwal: This has now been done ^^

src/app/views/forgotten-password/setForgottenPasswordController.jsx
dandwal: Like it, this has now been done

src/app/views/forgotten-password/setForgottenPasswordController.jsx
dandwal: Ah, so these two are actually storing two different things,
`passwordIsValid` is storing whether or not the password is valid, this is only used to track if we can submit the password to the server.
`showInputError` doesn't track whether the password is correct or not. It instead tracks if the user has attempted to submit the password before meeting the validation rules, and then tracks whether the user has corrected these issues.
To put it another way, 
- `passwordIsValid` is false from the start, `showInputError` is also false from the start.
- if the user types `passwordIsValid` may turn true, whereas `showInputError` remains as false
- If the user submits but the password is invalid, then `showInputError` will then turn true (so only after the user clicks the `Confirm password` button) It will then remain true until they fix all validation issues.

If we used just `passwordIsValid` then as soon as the page loads up the password field would be red, and this can cause issues in accessibility for 'anxious' users as we are immediately telling them something is wrong.

src/app/components/validation-item-list/ValidationItem.jsx
dandwal: I think Input of type checkboxes don't acknowledge the readonly attribute, however disabled is a good shout.

src/app/components/validation-item-list/ValidationItem.jsx
dandwal: Yer the onClick was there to capture clicks then discard them, as otherwise users could click the checkbox to change its state. With the disabled attribute added as per another comment, this is no longer needed so has been removed

src/app/components/validation-item-list/ValidationItem.jsx
AnnaEmson: it does as per W3C standard, which is readonly="readonly" on input. I checked it out. https://www.w3.org/TR/2010/WD-html5-20101019/the-input-element.html

src/app/views/forgotten-password/setForgottenPasswordController.jsx
AnnaEmson: Hi Daniel, if you fancy I can help you with this. you really do not need those two states. errors can be activated as a result of derived state from validation. But it is not big deal, just worth keeping the state as simple as possible.

src/app/components/validate-new-password/ValidateNewPassword.jsx
dandwal: I have cleaned up the state, however, we don't want to always show the error if they haven't met validation rules, only show it after a submit has happened (this is where the difference in the `passwordIsValid` and `showInputError` state vars come into play. I have added a gif to the PR showing how this works a little, sorry low quality (compression)

src/app/components/validation-item-list/ValidationItem.jsx
dandwal: The W3C standard in that link states: 

> The readonly attribute can also in some cases (e.g. for the Date state, but not the Checkbox state) make an input element immutable.

src/app/views/forgotten-password/setForgottenPasswordController.jsx
dandwal: Yes, I think it would be worth a conversation on this on Monday, if you are up for that :)

dandwal: Have moved the `passwordIsValid` to attribute in class rather than state var

src/app/components/validation-item-list/ValidationItem.jsx
dandwal: Since moving to a disabled state the onClick event has been removed

src/app/components/validation-item-list/ValidationItem.jsx
```diff
@@ -12,22 +12,11 @@ export default class ValidationItem extends Component {
...
-                <input
-                    type="checkbox"
-                    id={this.props.id}
-                    checked={this.props.checked}
-                    onClick={this.preventUserInteractions}
-                    onKeyDown={this.preventUserInteractions}
-                    className="readonly-checkbox"
-                />
-                <label htmlFor={this.props.id}> {this.props.name}</label>;
+                <input type="checkbox" id={this.props.id} checked={this.props.checked} disabled={"disabled"} className="readonly-checkbox" />
```
AnnaEmson, line 18:
```suggestion
                <input type="checkbox" id={this.props.id} checked={this.props.checked} disabled="disabled" className="readonly-checkbox" />
```
```suggestion
                <input type="checkbox" id={this.props.id} checked={this.props.checked} disabled={"disabled"} className="readonly-checkbox" />
```

src/app/views/forgotten-password/setForgottenPasswordController.jsx
```diff
@@ -2,22 +2,134 @@ import React, { Component } from "react";
...
+import user from "../../utilities/api-clients/user";
+import { errCodes } from "../../utilities/errorCodes";
+import notifications from "../../utilities/notifications";
+import log from "../../utilities/logging/log";
 
 const propTypes = {
     dispatch: PropTypes.func.isRequired
 };
 
+const status = {
```
AnnaEmson, line 14: loving this :-)

AnnaEmson (dismissed):
hi as we discussed, it would be nice to add those changes but as I am aware about timing we can always refactor later :-) so approving now

AnnaEmson (approved):
hi as we discussed, it would be nice to add those changes but as I am aware about timing we can always refactor later :-) so approving now

PR #581, opened 2021-09-13
Feature/class to function components

AnnaEmson:
### What

### How to review

### Who can review


src/app/App.jsx
```diff
@@ -1,55 +1,43 @@
-import React, { Component } from "react";
+import React, { Component, useEffect, useState } from "react";
```
JoshSinghONS, line 1: Am I right in thinking since this file is moving from a class component to a functional component we can stop importing Component?

src/app/App.jsx
```diff
@@ -1,55 +1,43 @@
...
-class App extends Component {
-    constructor(props) {
-        super(props);
-
-        this.state = {
-            isCheckingAuthentication: false
-        };
-    }
+const App = props => {
+    const [isCheckingAuthentication, setIsCheckingAuthentication] = useState(true);
```
JoshSinghONS, line 12: The old code initially set isCheckingAuthentication to false but the new code is setting it too true. Is this correct?

src/app/App.jsx
AnnaEmson: sure!

src/app/App.jsx
```diff
@@ -61,27 +49,28 @@ class App extends Component {
...
-    }
-}
+    return (
+        <div>
+            {isCheckingAuthentication && (
+                <div className="grid grid--align-center grid--align-self-center grid--full-height">
+                    <div className="loader loader--large loader--dark"></div>
+                </div>
+            )}
+            {props.children}
```
JoshSinghONS, line 63: Is the old code not saying if isCheckingAuthentication true then show div else show children, whilst the new is showing children regardless?

src/app/components/navbar/Navbar.jsx
```diff
@@ -1,54 +1,31 @@
...
+import url from "../../utilities/url";
+import cookies from "../../utilities/cookies";
+import auth from "../../utilities/auth";
+import user from "../../utilities/api-clients/user";
+import PreviewNav from "./preview-nav";
+
+const Navbar = ({ user, rootPath, workingOn, config, location }) => {
+    const authenticated = auth.isAuthenticated(user) || false;
+    const routeIsACollectionPage = path => {
+        return path.indexOf(`/collections`) >= 0;
```
JoshSinghONS, line 13: is path a object held natively in react?

src/app/components/navbar/Navbar.jsx
```diff
@@ -1,54 +1,31 @@
...
-        if (!showWorkingOn) {
-            return;
-        }
+    };
 
-        const path = this.props.location.pathname;
-        if (this.routeIsACollectionPage(path)) {
+    const renderWorkingOnItem = workingOn => {
+        if (!workingOn) return null;
+        if (routeIsACollectionPage(location.pathname)) {
```
JoshSinghONS, line 22: How are we able to use location? Does line 20 not need to be renderWorkingOnItem = (workingOn, location) => {

AnnaEmson: again, location is part of react router. it comes form wrapping those components in router :-)

PR #582, opened 2021-09-15
Feature/signin feature tests

mwnicks:
### What

Implemented sign in and sign out feature tests including support for old and new world auth.

### How to review

Ensure all the feature tests are passing

### Who can review

Anyone but me


features/steps/publisher.go
```diff
@@ -2,32 +2,58 @@ package steps
...
+	)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *Publisher) signOut() error {
```
dandwal, line 49: So sign out is actually currently broken.
We did signOut before sign in stories, since doing the signIn stories we decided to set cookies server-side for better security. However, since moving it to the backend it means that the signout functionality won't delete cookies as it attempts to do this with JS. So this feature test should be failing at the moment, until that is fixed.

features/steps/publisher.go
```diff
@@ -36,3 +62,58 @@ func (p *Publisher) logIn(username string) error {
 
 	return nil
 }
+
+func (p *Publisher) assignUserCookies() chromedp.Action {
```
dandwal, line 66: Might be my understanding but this is getting cookies right not assigning any cookies? The name might be a bit misleading.

features/user_sign_in.feature
```diff
@@ -0,0 +1,14 @@
+Feature: User sign in
+
+  As a Viewer, Publisher or Admin
+
+  Background:
+    Given I am not signed in
+
+  Scenario: A user signing in with valid credentials
```
dandwal, line 8: Still getting to grips with feature tests, should we be testing other cases, like when a user attempts to sign in but has not provided an email or password and that the screen shows the validation errors?

features/steps/publisher.go
dandwal: Although a little confused about how this is passing at the moment as we are setting the cookies in the test via `Given I have Auth tokens` then we are deleting the cookies in the codebase by `I sign out` but this would be being done with JS right, yet I can see the cookies in the test say `httpOnly`.

dandwal (approved):

PR #589, opened 2021-09-28
Legacy sign out updated

dandwal:
### What

As part of the new auth work, accounts are moving to a new backend and this aligns the legacy sign out functionality with that. We typically don't update the legacy code section, but as the new auth work is to do with security an exception was made for this.

Also, updated the Florence backend feature flag for endpoints which was nested under an extra feature flag. This meant that two feature flags had to be enabled - where just the ENABLE_NEW_SIGN_IN feature flag should have been required.

### How to review

Check the code, sign into Florence with your usual account, navigate to a legacy page (e.g. create a new collection then add content to it by clicking the 'old workspace') You know if you are on a legacy page if you see the 'signal' bars at the top left of the nav bar.

### Who can review

Anyone except me


mr-nick17 (approved):
LGTM

PR #598, opened 2021-10-07
Refactor class components to functional

AnnaEmson:
### What

This PR is still part of bigger task: Updating florence to React17

1. Bumping few earlier updated packages in react eco system to keep it updated. (sorry lots of changes made by prettier when updated)  
2.  Refactor App component to functional. 
3. Refactor Navbar and PreviewNav
4. Notifications and NotificationItem
5. Export NotFound to separate file with the vision to simplify the index file
6. Expand and add some unit tests. 

Refactoring to functional components is crucial to implement react-router-dom@5.3.0.
Why? Because new router works is based on hooks and hooks do not work on class components. we could, of course, find a solution to make wrapping component using hooks etc.. but we decided we would prefer refactor so we can also get rid of unsafe lifecycle methods.

### How to review

Check if everything is still working 

### Who can review

front end devs


src/app/components/preview-nav/PreviewNav.test.js
```diff
@@ -0,0 +1,113 @@
...
+});
+
+// TODO: when added react-router-dom to use mock location
+// it("routes to selected page on page selection", () => {
+//   component.handleSelectChange({ target: { value: "/test-uri" } });
+//   expect(dispatchedActions[1].type).toBe("@@router/CALL_HISTORY_METHOD");
+//   expect(dispatchedActions[1].payload.method).toBe("push");
+//   expect(dispatchedActions[1].payload.args[0]).toBe("/florence/collections/test-collection/preview?url=/test-uri");
+// });
+// });
```
mr-nick17, line 113: Please remove commented out code

AnnaEmson: Hi Nick, I left it by purpose as I need to revisit this when install new router but can remove it is a problem, just didn't want to spent time on something which I need to change in couple of weeks

mr-nick17: As this is lifted from another file and you've got a task to address this at a later date, I am OK with this commented out code being included

PR #599, opened 2021-10-11
On first time sign in hide 'forgotten password'

dandwal:
### What

Tiny addition, first time sign in user will come to Florence from an introductory email, this email contains a link with the URI paramater `first-time=true` if present / set to true then the link to 'forgotten password' should be hidden from the user

### How to review

Pull PR Run with feature flag `ENABLE_NEW_SIGN_IN=true` go to Florence login page `.../florence/login?first-time=true` there should be no link to forgotten password. Now do the same without the first-time=true (or set it to false) and the forgotten password link should be present 

### Who can review

Anyone except me


mike-kelly-methods (approved):
👍

PR #602, opened 2021-10-13
New Content Action Bar component

dandwal:
### What

A new component, 'Content Action Bar' which can be universally used across Florence. Note that there is a similar bar that is just used for editing dataset pages likewise there are similar looking items spread across Florence which are created inside of a page component (not as a component in its own right but a group of elements each time). This new component should be used going forward for all content action bars and potentially replace some of the others when pages get redesigned so we have some more continuity across Florence.

In the new User and Teams screens this bar will appear at the bottom of the screen to warn users there are unsaved changes and to allow them to save any changes.
e.g. 
Screenshot 2021-10-13 at 10 15 50

It was also requested that this component could be used on other content pages for a similar purpose and to be usable in place of the 'save and submit for review', 'preview', 'save' buttons etc. Some of these places actually use links that are disguised as buttons, so this is also possible to accomplish with this component if desired.

There is a new button style as per the UX designs which are used for the cancel button, this is an inverse of the primary button style colours with an additional border.

This component has been tested with Google lighthouse, VoiceOver and keyboard navigation. Also worth noting like, with any Florence item, Florence is only used on the network machines (windows, chrome, desktop).

Here are some examples of the component in use

Fixed to bottom
Screenshot 2021-10-12 at 15 54 29

Floating on page
Screenshot 2021-10-12 at 15 53 53

### How to review

Check the code
If desired add the component to a screen (`<ContentActionBar {...props} />` where props is an object whos available attributes can be found in the propTypes for the component (in the ContentActionBar.jsx file )

### Who can review

Anyone except me

### DIFF
--- a/src/app/views/login/SignIn.jsx
+++ b/src/app/views/login/SignIn.jsx
@@ -297,7 +297,7 @@ export class LoginController extends Component {
                 heading: "Change your password",
                 buttonText: "Change password",
                 requestPasswordChange: this.requestPasswordChange,
-                changeConformation: <ChangePasswordConfirmed handleClick={this.setPermissions} />,
+                changeConformation: <ChangePasswordConfirmed handleClick={this.setPermissions}/>,
                 status: this.state.status,
             };
 
@@ -329,35 +329,12 @@ export class LoginController extends Component {
                 },
             ];
             return (
-                <div>
-                    <LoginForm
-                        inputs={inputs}
-                        isSubmitting={this.state.status === status.SUBMITTING_SIGN_IN || this.state.status === status.SUBMITTING_PERMISSIONS}
-                        onSubmit={this.submitSignIn}
-                        validationErrors={this.validationErrors}
-                    />
-                    <ContentActionBar buttons={[{
-                        id: "meh",
-                        text: "Save content",
-                        interactionCallback: () => {
-                        },
-                        style: "positive",
-                        disabled: false
-                    },{
-                        id: "meh-2",
-                        text: "Preview",
-                        interactionCallback: () => {
-                        },
-                        style: "primary",
-                        disabled: false
-                    }]}
-                                      cancelCallback={() => {
-                                      }}
-                                      cancelDisabled={false}
-                                      stickToBottom={true}
-                                      unsavedChanges={true}
-                    />
-                </div>
+                <LoginForm
+                    inputs={inputs}
+                    isSubmitting={this.state.status === status.SUBMITTING_SIGN_IN || this.state.status === status.SUBMITTING_PERMISSIONS}
+                    onSubmit={this.submitSignIn}
+                    validationErrors={this.validationErrors}
+                />
             );
         }
     }


src/app/components/content-action-bar/ContentActionBar.jsx
```diff
@@ -0,0 +1,97 @@
...
+const buttonStyle = {
+    STANDARD: "standard",
+    POSITIVE: "positive",
+    PRIMARY: "primary",
+    WARNING: "warning",
+    SUBTLE: "subtle",
+    INVERT_PRIMARY: "invert-primary"
+};
+const buttonStyleArray = Object.values(buttonStyle);
+const propTypes = {
```
AnnaEmson, line 14: can we start using propTypes like recommended on the react docs https://reactjs.org/docs/typechecking-with-proptypes.html.
```
ContentActionBar.propTypes = {
  .... 
};
```

src/app/components/content-action-bar/ContentActionBar.jsx
```diff
@@ -0,0 +1,97 @@
+import React from "react";
+
+import PropTypes from "prop-types";
+
+const buttonStyle = {
```
AnnaEmson, line 5: I don't like those constants here. classNames should be passed as props if there will be different variants of the component

src/app/components/content-action-bar/ContentActionBar.jsx
```diff
@@ -0,0 +1,97 @@
...
+    cancelDisabled: PropTypes.bool,
+    stickToBottom: PropTypes.bool,
+    unsavedChanges: PropTypes.bool
+};
+
+const ContentActionBar = props => {
+    const outerContainerClasses = props.stickToBottom
+        ? "grid grid--justify-center content-action-bar content-action-bar--bottom-fixed"
+        : "content-action-bar";
+    const innerContainerClasses = props.stickToBottom ? "grid__col-10" : "";
```
AnnaEmson, line 35: here again, do you really need grid to display those buttons?

src/app/components/content-action-bar/ContentActionBar.jsx
```diff
@@ -0,0 +1,97 @@
...
+                            c9.22,0,16.696,7.475,16.696,16.696S265.22,422.05,256,422.05z M239.304,344.137V177.181h33.391v166.956H239.304z"
+                />
+            </svg>
+            <span className="content-action-bar__warn__text"> You have unsaved changes</span>
+        </span>
+    );
+    return (
+        <div className={outerContainerClasses}>
+            <div className={innerContainerClasses}>
+                <div className="padding-bottom--1 padding-top--1">
```
AnnaEmson, line 77: this is redundant you can dd those classes in the parent. I tried to find those classes in the repo , where do they reside?

src/app/components/content-action-bar/ContentActionBar.test.js
```diff
@@ -0,0 +1,75 @@
+import React from "react";
+import ContentActionBar from "./ContentActionBar.jsx";
+import renderer from "react-test-renderer";
+import { mount } from "enzyme";
+
+describe("ContentActionBar", () => {
+    describe("is created stuck at bottom with a single action button, cancel and with unsaved changes", () => {
```
AnnaEmson, line 7: describe usually should be used as a context description
like 
```suggestion
    describe("when props are passed ", () => {
```
I don't think it should be used in this condition. we have simple case here: only testing props being passed. I would remove it as it doesn't bring any value

src/app/components/content-action-bar/ContentActionBar.test.js
```diff
@@ -0,0 +1,75 @@
...
+            expect(component.find(".content-action-bar--bottom-fixed").length).toBe(1);
+            expect(component.find(".btn").length).toBe(2);
+            expect(component.find("#btn-cancel").exists()).toBe(true);
+            const fooButton = component.find("#foo-id");
+            expect(fooButton.exists()).toBe(true);
+            expect(fooButton.hasClass("btn--positive")).toEqual(true);
+            expect(component.find(".content-action-bar__warn").exists()).toBe(true);
+        });
+        const componentForSnapshot = renderer.create(<ContentActionBar {...contentActionBarProps} />);
+        expect(componentForSnapshot.toJSON()).toMatchSnapshot();
```
AnnaEmson, line 34: can you please extract this to separate test case. 
it("matches the snapshot", () => {

}

src/app/components/content-action-bar/ContentActionBar.jsx
dandwal: Yer I agree these can now be removed. Initially, there were a few classes being applied to various elements when I initially added them in, then optimised it a bit during further development but I didn't clock that this is actually redundant now. Good spot.

src/app/components/content-action-bar/ContentActionBar.test.js
dandwal: In regards to the testing comments, I don't think I quite follow, but have made some changes I hope are in line with your comments.

To me the description is to describe a unit of functionality and nested describes can explain a sub functionality of a testable unit. In this instance the unit of functionality was given a specific set of props does it function the way expected (the 'function' being does the component display the correct classes and elements). Then in the second test we are testing that given a change of different props it displays the expected elements and classes.

PR #603, opened 2021-10-15
bug: DatasetMetadata - fix errors in props

AnnaEmson:
### What

There are few errors in this component and this PR is solving them 

### How to review
check if you are able to create datasets correctly 
If you go to florence and log in, create a new collection (enter a name, choose manual publish and create), then choose create/edit content from the frame that comes up, choose filterable dataset and choose one from the list ('cantabular working' works), on the next page when you go to select a version (version 2 in this case) you'll be shown a blank page instead of the next step in the journey (edit metadata)

### Who can review


franmoore05 (approved):
LGTM

PR #609, opened 2021-10-21
Refactor/class components to func josh

JoshSinghONS:
### What

### How to review

### Who can review


src/app/components/layout/Layout.jsx
```diff
@@ -10,7 +10,7 @@ import NavBar from "../../components/navbar";
 
 const Layout = props => {
     const [isCheckingAuthentication, setIsCheckingAuthentication] = useState(null);
-
+
```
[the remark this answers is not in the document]
JoshSinghONS: This was, I've tried numerous times to change this back but have failed every time.

src/app/views/logs/Logs.jsx
```diff
@@ -143,63 +103,68 @@ class Logs extends Component {
...
-                        </div>
-                    )}
-                </div>
-            </div>
-        );
-    }
-}
+    const mapStateToProps = state => {
+        return {
+            page: state.routing.locationBeforeTransitions.query.page,
```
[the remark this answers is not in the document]
JoshSinghONS: Being honest I know little about stores but is it not referring to line 19 in store.js?

JoshSinghONS:
> Hi, I started doing review, good work but the code seems very strange. I wanted to investigate what response is coming form the server and I wasn't able to find this part of code being used anywhere in the app. when I follow routs it was not even working as there is no response coming form the server. did you verify if it was working before opened PR?

Hi. I did zero verification I kinda saw this as training piece as it was my first go at React work and saw this PR as a first draft more than anything. I also have no idea where one would begin to look at to see this code in action.

AnnaEmson:
Ok, so I think you would need to investigate if that functionality is being used as I think is not as it is broken even on develop. To see the component you refactored in action you need to go `http://localhost:8081/florence/logs` and follow the errors. We can  chat if you need more explanation.

src/app/views/logs/Logs.jsx
```diff
@@ -1,124 +1,83 @@
...
-            log.getAll((nextProps.page - 1) * 10, this.state.pageSize, this.state.logsTimestamp).then(logRange => {
-                this.setState({
-                    isFetchingLogs: false,
-                    logs: logRange,
-                });
+        } else {
+            setIsFetchingLogs(true);
+            log.getAll((props.page - 1) * 10, pageSize, logsTimestamp).then(logRange => {
+                setIsFetchingLogs(false);
+                setLogs(logRange);
```
JoshSinghONS, line 75: @AnnaEmson so looking at lines 63-75 seems to be the problem. Not quite sure I did the refactoring here correctly.

From my understanding I changed the UNSAFE function to the useEffect hook and reference the props said function is dependent on. However the code then didn't like that nextProps was undefined so I reworked the conditionals to what I think it should be as useEffect doesn't make use of nextProps and useEffect should only run when props.page changes.

However after inspecting the props when going to the logs route the page prop is reading as undefined and throwing the entire thing off.

Basically I'm stuck and above is what I was able to figure out. Any ideas?

AnnaEmson: so you confirmed the component `Logs.jsx` is used and you verified in production/ develop it is working?

AnnaEmson: so when I investigated this now on prod
I got `Request URL: https://publishing.develop.onsdigital.co.uk/zebedee/collectionDetails/logs
Request Method: GET
Status Code: 404 `
so it looks like those APIs were removed or changed. This need to be checked. If it got broken needs to be fixed or is not needed anymore and we could clear all that stuff.  

So this is why  I think it would be good of Product Manager to verify if this functionality is used?

JoshSinghONS: So I went to http://localhost:8081/florence/logs and kinda just tried to fix the errors the console log spat out to me.

JoshSinghONS: I understand what you mean though that if this is broke on prod I guess refactoring this file is moot?

PR #614, opened 2021-10-29
Feature/oldworld session timeout toasts

mwnicks:
### What

Added redirect paths to url when returning to login screen from legacy Florence on session timeout.
Added check on login endpoint to add session expired notification if a redirect is in the url to give users context.

### How to review

Log in to Florence.
Remove your session cookie.
Navigate to a legacy Florence page.
You should be taken to the login screen, the url of the page you tried to view should be in the url as a redirect query param, a toast notification should appear saying that your session has expired.

### Who can review

Anyone but me

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,7 +5,7 @@ import { Router, Route, IndexRoute, IndexRedirect, Redirect } from "react-router
 import { routerActions } from "react-router-redux";
 import { connectedReduxRedirect } from "redux-auth-wrapper/history3/redirect";
 import { store, history } from "./app/config/store";
-import { setConfig } from "./app/config/actions";
+import {setConfig} from "./app/config/actions";
 import auth from "./app/utilities/auth";
 import Layout from "./app/components/layout";
 import LoginController from "./app/views/login/LoginController";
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -45,7 +45,9 @@ import WorkflowPreview from "./app/views/workflow-preview/WorkflowPreview";
 import CreateContent from "./app/views/content/CreateContent";
 import NotFound from "./app/components/not-found";
 import "./scss/main.scss";
-
+import {errCodes} from "./app/utilities/errorCodes";
+import notifications from "./app/utilities/notifications";
+console.log('testing loading')
```
dandwal, line 50: May want to remove this line.

Not an issue but if removing the log line it may be worth adding back the spaces around the import braces on lines 8 and 48

dandwal (commented):
Just the one minor comment about console log. But other than that works as advertised, nice work ^^

src/index.js
mwnicks: Removed the console log and added the spaces in.

dandwal (approved):

PR #616, opened 2021-11-01
New popout component

dandwal:
### What

New popouts in Florence

There are currently modals in Florence which accept something of any shape. This new component makes use of the Modal but standardises what a popout should look like. Note that old world Florence has got some types of popouts called 'sweet alerts', react side Florence replaced most of these with toasts. However, the new auth work requires popouts as per UX designs.

In addition, the sessionManagement has been updated, so that when the timers expire a popout will appear.
The popouts have been configured in a similar way to the notifications and so can be triggered anywhere in Florence (react side of it)

Here are some examples of it in action:
Screenshot 2021-11-01 at 10 10 22
Screenshot 2021-11-01 at 10 11 17
Screenshot 2021-11-01 at 10 13 31

### How to review

Check out the code try make some popups if you want

#### How to create one
Find a file you would like to create a popout from.
Ensure the following are imported:

```JavaScript
import { store } from "../config/store";
import { addPopout, removePopouts } from "../config/actions";
```

Set some popoutOptions e.g.

```JavaScript
const popoutOptions = {
            id: "refresh-expire-soon",
            title: "Sorry, you need to sign back in again",
            body: "This is because you have been signed in for the maximum amount of time possible. Please save your work and sign back in to continue using Florence.",
            buttons: [
                {
                    onClick: () => {
                        store.dispatch(removePopouts(["refresh-expire-soon"]));
                    },
                    text: "Ok",
                    style: "primary",
                },
            ],
        };
        store.dispatch(addPopout(popoutOptions));
    };
```

Finally call your popout
```JavaScript
store.dispatch(addPopout(popoutOptions));
```

To remove it call

```Javascript
store.dispatch(removePopouts(["session-expire-soon", "refresh-expire-soon"]));
```

To remove it 

### Who can review

Anyone except me


mwnicks (dismissed):
LGTM 👍 
Except for the failing tests

mwnicks (approved):
LGTM 👍

PR #618, opened 2021-11-02
Legacy Florence session management added

dandwal:
### What

We don't typically update the legacy Florence side unless for security. The new Auth work is exactly that, we need to sort out the security around authentication and this new system should help that. As a result, there are some updates to the legacy system that are needed. Note this code will (hopefully) be one day removed once the entirety of the Legacy Florence code base has eventually be shifted over to react. In the meantime, this temporary solution is required in order to keep the lights on and the system secure.

#### So what is in this PR exactly

User session management. Sessions have typically been stored in Zebedee and users tokens were stored insecurely on the user's machine. This update is a step towards using a new Auth service which will be managed by our dp-idetntiy-api. Now users sign in and are given 3 tokens, 2 are relevant to this PR. These tokens are stored on the user's computer but in a secure manner now. When the user signs in they are given an access token and a refresh token. The access token is sent from the users computer to every outbound Florence request to say 'hey here is my proof that I am this uses' this token has a very short expiry time. The second token is used by dp-identity-api to get the user a new access token so that the user can stay signed in whilst they are using Florence without having to sign in every time their access token expires. 

When the client makes a request to the server to sign in, it provides these tokens but also the token expiry times. The expiry times are stored in local browser storage. This PR will make legacy Florence read the times these tokens will expire then set timers. The first timer monitors users interactions to see if they are still there. The second timer will create a popup (more invasive as the user could be reading content so not actively interacting but they are still there so we needed another way to check). The second two timers are similar but for the refresh token, this token can't be renewed so when the final timer goes off we ask the user to sign back in and we do one final renew to try give them as much time as possible to save their work and sign back in.

Why are we storing this in local browser session storage?  Well, Florence has two halves Legacy and React(new) the sign in pages are built in React but when the user goes to a legacy page it is a separate application. As such we need a way of transferring the expiry times between the two, the user can switch between old and new Florence pages through their time signed into Florence. 

### How to review

A lot of the backend work (around permissions) and frontend around team management and user management is ongoing so end to end tests are not possible at the moment, but you can test small snippets of functionality. Note, we have got tickets to test everything end to end once all parts of the journey are ready.

Probably the quickest way to demo this functionality is to alter the file `florence/src/legacy/js/functions/_setupFlorence.js`
Then remove the conditional part around the following lines:

```JavaScript
  if (Florence.globalVars.config.enableNewSignIn) {
      Florence.sessionManagement = {
          timers: {}
      };
      initialiseSessionExpiryTimers();
  }
```

Then alter the file `florence/src/legacy/js/functions/_sessionManagement.js` and add 

```JavaScript
  let mockSET = new Date();
  let mockRET = new Date();
  mockSET.setSeconds(mockSET.getSeconds() + 20) // Change 20 to however many seconds you want to wait for the timers to go off
  mockRET.setMinutes(mockRET.getMinutes() + 1) // Change 1 to however many minutes you want to wait for the timers to go off
  mockSET = mockSET.toISOString();
  mockRET = mockRET.toISOString();
  mockSET += " +0000 UTC"
  mockRET += " +0000 UTC"
  sessionStorage.setItem("session_expiry_time", mockSET);
  sessionStorage.setItem("refresh_expiry_time", mockRET);
```

Then change the following set of lines in the file

```JavaScript
startExpiryTimer("sessionTimerPassive", sessionExpiryTime, 300000, monitorInteraction);
startExpiryTimer("sessionTimerInvasive", sessionExpiryTime, 120000, warnSessionSoonExpire);
startExpiryTimer("refreshTimerPassive", refreshExpiryTime, 300000, monitorInteraction);
startExpiryTimer("refreshTimerInvasive", refreshExpiryTime, 120000, warnRefreshSoonExpire);
```

To

```JavaScript
startExpiryTimer("sessionTimerPassive", sessionExpiryTime, 10000, monitorInteraction);
startExpiryTimer("sessionTimerInvasive", sessionExpiryTime, 0, warnSessionSoonExpire);
startExpiryTimer("refreshTimerPassive", refreshExpiryTime, 10000, monitorInteraction);
startExpiryTimer("refreshTimerInvasive", refreshExpiryTime, 0, warnRefreshSoonExpire);
```

You will then need to rebuild, to do this run `make node-modules` then run the usual `make debug` command

Note you will get an error saying something like can't keep you signed in, this is because your using a Zebedee account and trying to refresh the access token on a dp-identity-api (cognito) account which doesn't exist there.

It is also worth noting the response body leaving Florence from eh frontend to the backend of Florence is empty. This is because the token needed to refresh the access token is actually added at the backed Florence side (as this is a httpOnly cookie JS can't access it - as per design) so only server-side can access this cookie

### Who can review

Anyone except me

### DIFF
--- a/src/app/components/layout/Layout.jsx
+++ b/src/app/components/layout/Layout.jsx
@@ -7,6 +7,7 @@ import sessionManagement from "../../utilities/sessionManagement";
 import user from "../../utilities/api-clients/user";
 import Notifications from "../notifications";
 import NavBar from "../../components/navbar";
+import Popouts from "../popouts/Popouts";
 
 const Layout = props => {
     const [isCheckingAuthentication, setIsCheckingAuthentication] = useState(null);
@@ -68,6 +69,7 @@ const Layout = props => {
             <NavBar location={props.location} />
             {props.children}
             {props.notifications && <Notifications notifications={props.notifications} />}
+            {props.popouts && <Popouts popouts={props.popouts} />}
         </div>
     );
 };
--- a/src/app/components/layout/index.jsx
+++ b/src/app/components/layout/index.jsx
@@ -4,6 +4,7 @@ import Layout from "./Layout";
 function mapStateToProps(state) {
     return {
         notifications: state.state.notifications,
+        popouts: state.state.popouts,
     };
 }
 
--- a/src/app/components/popouts/PopoutItem.jsx
+++ b/src/app/components/popouts/PopoutItem.jsx
@@ -1,0 +1,42 @@
+import React from "react";
+import PropTypes from "prop-types";
+import clsx from "clsx";
+
+const propTypes = {
+    id: PropTypes.string.isRequired,
+    title: PropTypes.string,
+    body: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+    buttons: PropTypes.arrayOf(
+        PropTypes.shape({
+            onClick: PropTypes.func.isRequired,
+            text: PropTypes.string.isRequired,
+            style: PropTypes.oneOf(["standard", "positive", "primary", "warning", "subtle", "invert-primary"]),
+        })
+    ),
+};
+
+const PopoutItem = props => (
+    <div>
+        <div className="modal__header">
+            <h1 className="modal__title">{props.title}</h1>
+        </div>
+        <div className="modal__body">
+            {props.body}
+            <div className="modal__button-container">
+                {props.buttons.map((button, i) => {
+                    const buttonClasses = clsx("btn", "btn--margin-right", {
+                        [`btn--${button.style}`]: button.style && button.style !== "standard",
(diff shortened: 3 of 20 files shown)


AnnaEmson:
if not ready for review yet open as a draft :-)

dandwal:
> if not ready for review yet open as a draft :-)

It was a draft, then made it 'ready' but left the commend in the description saying 'not ready' which I should have removed. Taken to added the comment as have had people review the PR in draft state before. 

Looks like Concourse had errors running it didn't manage to execute the build, audit or unit, triggering these now so hopefully they will run this time.

mwnicks (commented):
Changes look good to me but couldn't reproduce changes locally, I think an issue with my local env rather than the changes.

mwnicks (approved):
Changes demoed and are working as expected. 👍

PR #621, opened 2021-11-04
Feature/simple selectable table

mwnicks:
### What

Updated the simple selectable list component to support optional additional columns

### How to review

- Ensure all tests are passing

- Ensure the existing tables still render correctly an example can be found at http://localhost:8081/florence/collections/< collection id >/create.

- Create/add to a list that contains an 'extraDetails' field. The additional columns should render from left to right as in the designs on this [ticket](https://trello.com/c/1CGZFLp9/558-update-simple-selectable-list-component-in-florence-to-new-design)

### Who can review

Anyone but me

### DIFF
--- a/src/app/components/simple-selectable-list/SimpleSelectableListItem.jsx
+++ b/src/app/components/simple-selectable-list/SimpleSelectableListItem.jsx
@@ -8,7 +8,7 @@ const propTypes = {
     url: PropTypes.string.isRequired,
     externalLink: PropTypes.bool.isRequired,
     details: PropTypes.arrayOf(PropTypes.string),
-    extraDetails: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.string)),
+    extraDetails: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.shape({content:PropTypes.string, classes:PropTypes.string}))),
     colCount: PropTypes.number.isRequired,
     disabled: PropTypes.bool,
 };
--- a/src/app/components/simple-selectable-list/SimpleSelectableListItem.jsx
+++ b/src/app/components/simple-selectable-list/SimpleSelectableListItem.jsx
@@ -56,7 +56,7 @@ export default class SimpleSelectableListItem extends Component {
                 </div>
                 {extraDetails.map((column, i) => {
                     return (
-                        <div key={`detail-${i}`} className={`simple-select-list__cols-${colCount}`}>
+                        <div key={`detail-${i}`} className={`simple-select-list__col simple-select-list__cols-${colCount}`}>
                             {column.map((detail, j) => {
                                 return (
                                     <span key={`detail-${j}`} className={detail.classes}>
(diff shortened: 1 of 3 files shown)


dandwal (commented):
Review is currently WIP will continue it tomorrow

I can see the current style has underlines for links but the updated styles do not. Are we moving towards removing them for this component or is it fine to leave the underlines in?

src/scss/components/_simple-select-list.scss
```diff
@@ -24,6 +26,43 @@
...
+            text-align: right;
+        }
+
+        &:first-of-type {
+            text-align: left;
+        }
+    }
+
+    //To keep the column size consistent between rows in the list.
+    &__cols {
```
dandwal, line 40: I think this is forcing the flexbox to act like a grid, it may be a better shout to use grid rather than flexbox if needing to achieve this. This would mean a new component if we are aiming to keep all columns even width across all rows as I think this component wasn't initially designed to be a table but just a simple list (as per the name). Aware this would inflate this task considerably.

From looking at the designs though, I think we can get away with still using this component but it doesn't need to have even columns across the rows. So maybe we can get away with removing this functionality and the colCount logic? It is possible that I have missed a use case though if so happy to discuss it further.

mwnicks: It doesn't explicitly say the columns need to be evenly sized on the ticket. I was basing it on the design attached to the ticket which shows the user 'chips' aligned on the left between the rows.
Happy to take this out if we don't think its needed, I wasn't happy with the code anyway.

src/app/components/simple-selectable-list/SimpleSelectableListItem.jsx
```diff
@@ -42,11 +44,28 @@ export default class SimpleSelectableListItem extends Component {
...
+                        return <p key={`detail-${i}`}>{detail}</p>;
+                    })}
+                </div>
+                {extraDetails.map((column, i) => {
+                    return (
+                        <div key={`detail-${i}`} className={`simple-select-list__col simple-select-list__cols-${colCount}`}>
+                            {column.map((detail, j) => {
+                                return (
+                                    <span key={`detail-${j}`} className={detail.classes}>
+                                        {detail.content}
```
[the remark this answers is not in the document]
mwnicks: The reason I went for a 2D array mapping to individual spans was to be able to support things like the 'chip' icons where it felt like they needed to individual elements rather than a combined string.

src/app/components/simple-selectable-list/SimpleSelectableListItem.jsx
```diff
@@ -42,11 +44,28 @@ export default class SimpleSelectableListItem extends Component {
...
-                {this.renderTitle()}
-                {details.map((detail, i) => {
-                    return <p key={`detail-${i}`}>{detail}</p>;
+                <div className={`simple-select-list__col simple-select-list__cols-${colCount}`}>
+                    {this.renderTitle()}
+                    {details.map((detail, i) => {
+                        return <p key={`detail-${i}`}>{detail}</p>;
+                    })}
+                </div>
+                {extraDetails.map((column, i) => {
```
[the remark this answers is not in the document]
mwnicks: Yeah, this does need updating. Will do once data structure is confirmed.

src/scss/components/_simple-select-list.scss
dandwal: Going to resolve this thread as we have discussed and we are going to continue with this current implementation as the work has been done

dandwal (approved):

PR #622, opened 2021-11-08
feature: Resolve Collection Creation Error Message

AnnaEmson:
### What
It was reported to us the users doesn't get proper form validation on name.  
Acceptance Criteria: `They system checks that the Collection name does not already exist before creating the collection (rather than creating the collection before checking the existence of the collection name)` 
 Ticket: [Resolve Collection Creation Error Message](https://trello.com/c/urvOoUjO/563-resolve-collection-creation-error-message)
ezgif com-gif-maker (2)

Users reported that even unique names are reported to be 'taken'. I wasn't able to verify it. 
### How to review

Check the code if doesn't contain any logical errors and meet the coding standards. 

### Who can review

Front end dev

### DIFF
--- a/src/app/views/collections/create/CollectionCreateController.jsx
+++ b/src/app/views/collections/create/CollectionCreateController.jsx
@@ -1,7 +1,6 @@
 import React, { Component } from "react";
 import { connect } from "react-redux";
 import PropTypes from "prop-types";
-
 import collections from "../../../utilities/api-clients/collections";
 import teams from "../../../utilities/api-clients/teams";
 import notifications from "../../../utilities/notifications";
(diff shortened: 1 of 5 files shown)


src/app/views/collections/create/CollectionCreateController.jsx
```diff
@@ -132,6 +131,23 @@ export class CollectionCreateController extends Component {
...
+        if (nameTaken) {
+            const collectionName = {
+                value: name,
+                errorMsg: "A collection with this name already exists",
+            };
+            const newCollectionDetails = {
+                ...this.state.newCollectionDetails,
+                name: collectionName,
+            };
+            this.setState({ newCollectionDetails });
```
JoshSinghONS, line 147: A bit confused how this comes out to show the error. 

Is there code somewhere which essentially checks for state.newCollectionDetails.errorMsg and if true output the error?

JoshSinghONS:
I also have no idea how to review src/app/views/collections/create/__snapshots__/CollectionCreateController.test.js.snap

src/app/views/collections/create/CollectionCreateController.jsx
AnnaEmson: Hi @JoshSinghONS , thank you for looks. It would be helpful to leave the mark on the channel so others know you are looking. 
Yes, this all happens in a child component on here: https://github.com/ONSdigital/florence/blob/develop/src/app/views/collections/create/CollectionCreate.jsx#L105

JoshSinghONS (approved):

PR #624, opened 2021-11-10
New component TextArea

dandwal:
### What

There is already a text area component buried in the `<Input>` component but it doesn't quite meet the needs of the new Auth screens in addition it doesn't meet the new UX for the component. Because of this, we would need to make quite a few adaptations and add more options to it. The `<Input>` component which is the container for the current textarea component is already a bit of a monolithic component. 

I don't think `<Input>` should be handling HTML textarea as well as HTML input as this is what is causing the convoluted mess in the component. As such this PR introduces a clean new text area component that can meet the new UX designs without breaking current text areas in play.

At some point, a small addition to the session management work had been accidentally dropped. This adds it into the codebase specifically around first time sign in, when the user changes their password they don't need to sign in again, as such they are given the expiration times of their auth tokens on a first time sign in password change success. This needs to be read so that the session management timers can start.

__Note__ this has been tested with keyboard navigation and with a screen reader, specifically the 'Mac VoiceOver Utility'.

### How to review

Check out the code, make sure best practices are being followed
If you need a hand or a demo remember to reach out, happy to go through it.

Here is a vid example (note the button is just for demo purposes)
Screen Recording 2021-11-10 at 11 42 24
Better quality vid can be found here: https://user-images.githubusercontent.com/47502916/141109253-ea684860-0d9e-4c4c-96a0-a059ab06e5b0.mov 

If you would like to try it out for yourself here is an example you can throw on any screen (change the props as you see fit).

```JavaScript 
        const [demoText, setDemoText] = useState("");
    const [errorText, setErrorText] = useState("");
    return (
            <div className="grid__col-4">
                <form className="form margin-top--5">
                    <TextArea
                        title={"Please provide some feedback"}
                        name={"feedback"}
                        text={demoText}
                        textChange={setDemoText}
                        descriptionHint={"For example, describe any difficulties you experienced in the use of this service"}
                        maxLength={200}
                        placeholder={"Please provide your feedback"}
                        rows={8}
                        error={errorText}
                    />
                </form>
                <button onClick={e =>{setErrorText("Something went wrong")}}>Make it error</button>
            </div>
        </div>
    );
```

 Remember to import the component and useState it to whatever screen you are adding it to e.g. if using the sign in screen

```JavaScript
import React, {useState} from "react";
import TextArea from "../../components/text-area/TextArea";

```

### Who can review

Anyone except me

### DIFF
--- a/src/app/components/text-area/TextArea.jsx
+++ b/src/app/components/text-area/TextArea.jsx
@@ -18,29 +18,37 @@ const propTypes = {
 
 const TextArea = props => {
     const charactersRemaining = props.maxLength - props.text.length;
-    const formClasses = clsx("form__input", {"form__input--error": props.error});
-    const textAreaBody =
+    const formClasses = clsx("form__input", { "form__input--error": props.error });
+    const textAreaBody = (
         <div className={formClasses}>
-            <label htmlFor={props.name} className="form__label">{props.title}</label>
+            <label htmlFor={props.name} className="form__label">
+                {props.title}
+            </label>
             <div>{props.descriptionHint}</div>
-            <textarea id={props.name} name={props.name} className="input input__textarea" value={props.text}
-                      onChange={e => props.textChange(e.target.value)} placeholder={props.placeholder}
-                      aria-describedby={`${props.name}-char-limit-remaining`} rows={props.rows}
-                      maxLength={props.maxLength}/>
+            <textarea
+                id={props.name}
+                name={props.name}
+                className="input input__textarea"
+                value={props.text}
+                onChange={e => props.textChange(e.target.value)}
+                placeholder={props.placeholder}
+                aria-describedby={`${props.name}-char-limit-remaining`}
+                rows={props.rows}
+                maxLength={props.maxLength}
+            />
             {props.maxLength && (
                 <div id={`${props.name}-char-limit-remaining`}>
                     <b>You have {charactersRemaining} characters remaining</b>
                 </div>
             )}
         </div>
+    );
 
     // If there is an error then render the text-area within an error panel
     if (props.error) {
-        return <Panel type="error"
-                      heading={props.error}
-                      body={textAreaBody}/>
+        return <Panel type="error" heading={props.error} body={textAreaBody} />;
     }
-    return textAreaBody
+    return textAreaBody;
 };
 
 const defaultProps = {
--- a/src/app/components/text-area/TextArea.jsx
+++ b/src/app/components/text-area/TextArea.jsx
@@ -1,0 +1,53 @@
+import React from "react";
+
+import PropTypes from "prop-types";
+import clsx from "clsx";
+import Panel from "../panel/Panel";
+
+const propTypes = {
+    name: PropTypes.string.isRequired,
+    text: PropTypes.string.isRequired,
+    textChange: PropTypes.func.isRequired,
+    title: PropTypes.string.isRequired,
+    descriptionHint: PropTypes.string,
+    error: PropTypes.string,
+    maxLength: PropTypes.number,
+    placeholder: PropTypes.string,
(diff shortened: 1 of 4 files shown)


src/app/components/text-area/TextArea.jsx
```diff
@@ -0,0 +1,61 @@
+import React from "react";
+
```
AnnaEmson, line 2: remove this line

src/app/components/text-area/TextArea.jsx
```diff
@@ -0,0 +1,61 @@
...
+    descriptionHint: PropTypes.string,
+    error: PropTypes.string,
+    maxLength: PropTypes.number,
+    placeholder: PropTypes.string,
+    rows: PropTypes.number,
+};
+
+const TextArea = props => {
+    const charactersRemaining = props.maxLength - props.text.length;
+    const formClasses = clsx("form__input", { "form__input--error": props.error });
```
AnnaEmson, line 21: can you please move it directly to the className, no need for this const

src/app/components/text-area/TextArea.jsx
```diff
@@ -0,0 +1,61 @@
...
+                value={props.text}
+                onChange={e => props.textChange(e.target.value)}
+                placeholder={props.placeholder}
+                aria-describedby={`${props.name}-char-limit-remaining`}
+                rows={props.rows}
+                maxLength={props.maxLength}
+            />
+            {props.maxLength && (
+                <div id={`${props.name}-char-limit-remaining`}>
+                    <b>You have {charactersRemaining} characters remaining</b>
```
AnnaEmson, line 41: <b> tag is obsolete, can you please use <strong>

src/app/components/text-area/TextArea.jsx
```diff
@@ -0,0 +1,61 @@
+import React from "react";
+
+import PropTypes from "prop-types";
+import clsx from "clsx";
+import Panel from "../panel/Panel";
+
+const propTypes = {
+    name: PropTypes.string.isRequired,
+    text: PropTypes.string.isRequired,
+    textChange: PropTypes.func.isRequired,
```
AnnaEmson, line 10: `handleChange` ?

AnnaEmson (approved):
nice work, just little improvements

src/app/components/text-area/TextArea.jsx
```diff
@@ -0,0 +1,61 @@
...
+            <label htmlFor={props.name} className="form__label">
+                {props.title}
+            </label>
+            <div>{props.descriptionHint}</div>
+            <textarea
+                id={props.name}
+                name={props.name}
+                className="input input__textarea"
+                value={props.text}
+                onChange={e => props.textChange(e.target.value)}
```
AnnaEmson, line 33: I like placing all actions at the end of props :-)

src/app/components/text-area/TextArea.jsx
```diff
@@ -0,0 +1,61 @@
...
+    }
+    return textAreaBody;
+};
+
+const defaultProps = {
+    rows: 8,
+};
+
+TextArea.propTypes = propTypes;
+TextArea.defaultProps = defaultProps;
```
AnnaEmson, line 59: `defaultProps on functional components will eventually be deprecated (as per Dan Abramov, one of the core team), so for future-proofing it's worth using default parameters.`

dandwal: Ah this is good to know in which case I will axe this thanks

src/app/components/text-area/TextArea.jsx
AnnaEmson: oh sorry I haven't noticed strong disappeared and GitHub formatted my note instead lol

AnnaEmson (approved):

PR #626, opened 2021-11-11
Feature/new world users screen

mwnicks:
### What

Create new auth world Users screen as per designs on [ticket](https://trello.com/c/sqRBKWxA/559-create-the-list-users-screen)

### How to review

(You will need florence, dp-api-router, zebedee and dp-identity-api running to test these changes)

The easiest way to test whilst the new auth is still being implemented is to make the following changes locally:
src/index.js line 146
`<Route path={`${rootPath}/users`} component={userIsAuthenticated(userIsAdminOrEditor(!config.enableNewSignIn ? UsersListController : UsersController))}>`

Then log in to Florence
Go to `/florence/users`.
The users should be listed as per designs in the above ticket.

### Who can review

Anyone but me


src/app/views/users/UsersListController.jsx
```diff
@@ -0,0 +1,183 @@
...
+    dispatch: PropTypes.func.isRequired,
+    children: PropTypes.element,
+    params: PropTypes.shape({
+        userID: PropTypes.string,
+    }).isRequired,
+    loggedInUser: PropTypes.object,
+    users: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
+};
+
+export class UsersListController extends Component {
```
AnnaEmson, line 27: Hi Nick, I can see it is in draft still, but just to give you heads up. It would be great if we start avoiding classes where possible and use functional components instead. Also, this file is no way any controller just UI, can you please name it `UsersList`. Please also make use of store.

mwnicks: Thanks Anna, I'm addressing these comment.

src/app/utilities/api-clients/user.js
```diff
@@ -12,10 +12,14 @@ export default class user {
         return http.get(`/zebedee/users?email=${email}`);
     }
 
-    static getAll() {
+    static getAllZebedee() {
```
dandwal, line 15: Not a change request, but think it may be safer to expand the getAll which was there and put a conditional in so we don't need to worry about making sure we renamed all calls, there is a slight risk some file has imported the 'user' file with a different alias which may mean a 'find all' may miss a few.

e.g.
```JavaScript
static getAll() {
        const config = window.getEnv();
        if (config.enableNewSignIn) {
            return http.get(`/users`, true);
        }
        return http.get(`/zebedee/users`);
    }
```

Also by referencing the feature flag it should make it easier at the end when we clean up and delete the getAllZebedee we do actually delete it as we are likely to do a code base search for the feature flag.

src/app/views/users/UsersList.jsx
```diff
@@ -0,0 +1,174 @@
...
+    params: PropTypes.shape({
+        userID: PropTypes.string,
+    }).isRequired,
+    loggedInUser: PropTypes.object,
+    users: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
+};
+
+export const UsersList = (props) => {
+    const [isFetchingUsers, setIsFetchingUsers] = useState(false);
+    const isAdmin = auth.isAdmin(props.loggedInUser);
```
dandwal, line 120: Haven't run the code, but just a double-check does this work if you sign in then in a new tab go straight to this screen? I remember there were issues around checking isAdmin on load and redirecting as a result. useEffect is called after first renderer. But, I don't know if the `auth` would have been updated in time as it makes a request to permissions or something and if not returned in time the user will be kicked out. I think if the user was always coming from the collections screen it should work but if first load on react side of Florence I am not 100% sure. Before the upgrade this was done in the component will mount lifecycle function, which was called before first render, but we don't have that available anymore.

src/app/views/users/UsersList.jsx
```diff
@@ -0,0 +1,174 @@
...
+                    </div>
+                    <div className="grid">
+                        <div className="grid__col-4">
+                            <Input id="search-content-types" placeholder="Search user by name or email" onChange={handleSearchInput} />
+                        </div>
+                    </div>
+                    <SimpleSelectableList rows={filteredUsers ? filteredUsers : props.users} showLoadingState={isFetchingUsers} />
+                </div>
+            </div>
+            {props.children}
```
dandwal, line 159: Do we need this, couldn't see it in the propTypes so not sure what this is doing - if not we do remove it then the outer div can go too I think.

src/app/views/users/UsersList.jsx
```diff
@@ -0,0 +1,174 @@
...
+            autoDismiss: 3000,
+        };
+        notifications.add(notification);
+        console.error("Error mapping users to state: ", error);
+        log.event("Error mapping users to state", log.error(error));
+        return false;
+    }
+};
+
+const propTypes = {
```
dandwal, line 107: Could we move these to the top of file under the imports, as we haven't got documentation on what components exist and how to use them the easiest way for someone to see how to use a component is to check the propTypes, so think it is better to be at the top of the file to save scrolling.

src/app/views/users/UsersList.jsx
```diff
@@ -0,0 +1,174 @@
...
+    rootPath: PropTypes.string.isRequired,
+    dispatch: PropTypes.func.isRequired,
+    children: PropTypes.element,
+    params: PropTypes.shape({
+        userID: PropTypes.string,
+    }).isRequired,
+    loggedInUser: PropTypes.object,
+    users: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
```
dandwal, line 115: I think we know the fields of the users object, so we should prob populate the shape.

src/app/views/users/UsersList.jsx
```diff
@@ -0,0 +1,174 @@
...
+import log from "../../utilities/logging/log";
+import auth from "../../utilities/auth";
+
+import { addAllUsers } from "../../config/actions";
+import SimpleSelectableList from "../../components/simple-selectable-list/SimpleSelectableList";
+import BackButton from "../../components/button/BackButton";
+import Input from "../../components/Input";
+
+
+export const getAllUsers = (dispatch, rootPath, setIsFetchingUsers) => {
```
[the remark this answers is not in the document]
mwnicks: It is exported so it can be called in the tests.

src/app/views/users/UsersList.jsx
mwnicks: I have tried the steps you described. When I go straight to this screen in a new tab the page does load correctly.

src/app/utilities/api-clients/user.js
```diff
@@ -12,10 +12,14 @@ export default class user {
...
         return http.get(`/zebedee/users`);
     }
 
+    static getAll() {
+        return http.get(`/users`);
```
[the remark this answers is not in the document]
mwnicks: Have added a param object that can be passed in and parsed into a query string to support filtering the list in the future.

src/app/views/users/UsersList.jsx
```diff
@@ -0,0 +1,174 @@
+import React, {Component, useEffect, useState} from "react";
```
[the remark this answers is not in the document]
mwnicks: Removed

src/app/views/users/UsersList.jsx
```diff
@@ -0,0 +1,174 @@
...
+import { connect } from "react-redux";
+import { push, replace } from "react-router-redux";
```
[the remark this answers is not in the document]
mwnicks: Removed

src/app/views/users/UsersList.jsx
```diff
@@ -0,0 +1,174 @@
...
+import BackButton from "../../components/button/BackButton";
+import Input from "../../components/Input";
+
+
+export const getAllUsers = (dispatch, rootPath, setIsFetchingUsers) => {
+    setIsFetchingUsers(true);
+    users
+      .getAll()
+      .then(allUsersResponse => {
+          const allUsers = allUsersResponse.users.map(user => {
```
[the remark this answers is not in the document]
mwnicks: I'm still a bit unsure about when we should be dispatching to store too.

src/app/views/users/UsersList.jsx
```diff
@@ -0,0 +1,174 @@
...
+    users: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
+};
+
+export const UsersList = (props) => {
+    const [isFetchingUsers, setIsFetchingUsers] = useState(false);
+    const isAdmin = auth.isAdmin(props.loggedInUser);
+    const [filteredUsers, setFilteredUsers] = useState(null);
+
+    useEffect(() => {
+        if (!isAdmin && !props.params.userID) {
```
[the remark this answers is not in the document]
mwnicks: I think that is a question for the James'

src/app/views/users/UsersList.jsx
```diff
@@ -0,0 +1,174 @@
...
+            user => user.title.toLowerCase().search(searchTerm) !== -1 || user.email.toLowerCase().search(searchTerm) !== -1
+        );
+        setFilteredUsers(filteredUsers);
+    };
+
+    return (
+        <div>
+            <div className="grid grid--justify-space-around">
+                <div className="grid__col-9">
+                    <BackButton fill="#000000" url={`${props.rootPath}/collections`} />
```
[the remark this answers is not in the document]
mwnicks: Removed the new component and switched to this.

src/app/views/users/UsersList.jsx
```diff
@@ -0,0 +1,174 @@
...
+                    <div className="grid grid--align-baseline">
+                        <div className="grid__col-1">
+                            <h1>Users</h1>
+                        </div>
+                        <div className="grid__col-1">
+                            <a href="#">Create new user</a>
```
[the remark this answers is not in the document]
mwnicks: URL added

src/app/utilities/api-clients/user.js
mwnicks: I tried this initially and it caused failures in other places. Looks like it was the positioning of the config initialisation that was the problem. Updated now

src/app/views/users/UsersList.jsx
mwnicks: Missed this, was left from the UsersController should have been taken out.

src/app/views/users/UsersList.jsx
mwnicks: Moved

src/app/views/users/UsersList.jsx
mwnicks: Fields added.

src/app/views/users/UsersList.jsx
```diff
@@ -0,0 +1,174 @@
...
+          setIsFetchingUsers(false);
+      })
+      .catch(error => {
+          setIsFetchingUsers(false);
+          log.event("Error fetching users", log.data({ status_code: error.status }), log.error(error));
+          switch (error) {
+              case 404: {
+                  const notification = {
+                      type: "warning",
+                      message: `No API route available to get users.`,
```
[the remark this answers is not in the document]
mwnicks: I think I've understood what you are saying here and have made the change.

mwnicks:
@AnnaEmson Dan suggested asking you to review this as well.

dandwal (approved):

PR #629, opened 2021-11-15
New component Dynamic List item

dandwal:
### What

This is another new component used in the UX designs for the new Auth work, I believe this is the last one needed. At current, this is planned to be used on a screen for adding members (who are neither publishers nor admins) to 'teams' also referred to as 'groups'.

This component should be accessible and has been tested with keyboard navigation as well at VoiceOver and the usual Lighthouse/Axe.

It can hold two articles of text and an action button. e.g. in this example when the user would click 'Add' next to a name, they would be removed from the list. In addition, this list has a search field that filters the list. When there are no results found in the list text can be shown instead.

It is worth noting this component is a list, not a table. It is also the responsibility of the parent component to order the list however desired.

See how it looks in action:
dynamic-list

A better resolution video can be found here:
https://user-images.githubusercontent.com/47502916/141754759-0cf69dba-8482-4f4c-8b70-3bed8154aba6.mov

### How to review

Check the code is it clean?
Optionally try the code out and try to make use of the component. If preferred reach out and I can demo it.

If you do want to try it out yourself here is an example of how

```JavaScript
const filteredList = [
        {
            title: "Simon Davies",
            desc: " Simon.davies@ons.gov.uk",
            icon: "Person",
            iconName: "Viewer",
            buttonName: "Add",
            buttonCallback: () => {},
        },
        {
            title: "Michael Miller",
            desc: " Michael.miller@ons.gov.uk",
            icon: "Person",
            iconName: "Viewer",
            buttonName: "Add",
            buttonCallback: () => {},
        },
    ];

const dynamicListProps = {
            title: "Add member to team",
            listItems: filteredList,
            listHeightClass: "fifth-height",
            noResultsText: "No viewers found",
            filterPlaceholder: "Search viewers by name or email",
            handleSearchInput: () => {console.log("I was pressed")},
<DynamicList {...dynamicListProps} />
```

Then optionally add some CSS to set the height (to make it scrollable rather than being endless e.g.

```scss
.fifth-height {
  height: 25vh
}
```

### Who can review

Anyone except me


src/scss/components/_dynamic_list.scss
```diff
@@ -0,0 +1,56 @@
...
+
+  &__container {
+    overflow: auto;
+  }
+}
+
+.dynamic-list-item {
+  display: flex;
+  list-style: none;
+  padding: $magicnumber1;
```
rav-pradhan, line 16: What does `magicnumber1` refer to? Could it be renamed to something more meaningful? 😄

src/scss/components/_dynamic_list.scss
```diff
@@ -0,0 +1,56 @@
...
+    padding: $magicnumber1;
+    background-color: $grey5;
+
+    &:focus {
+      outline: 3px solid $focus;
+    }
+  }
+
+  &__separator {
+    margin: auto $magicnumber1 auto $magicnumber1;
```
rav-pradhan, line 53: We can simplify this by having `margin: auto $magicnumber1` - this shorthand should set top and bottom margins to `auto`, and left and right margins to the value of `magicnumber1`

src/scss/components/_dynamic_list.scss
```diff
@@ -0,0 +1,56 @@
+.dynamic-list {
+  &__title {
+    font-size: 21px;
```
rav-pradhan, line 3: I'm guessing `21px` is common in other parts of Florence for headings etc - do we not have styling/classes in places to cover these typographies more broadly?

src/app/components/dynamic-list/DynamicList.jsx
```diff
@@ -0,0 +1,46 @@
...
+    if (props.listItems.length > 0) {
+        list = (
+            <ul className={`dynamic-list__container ${props.listHeightClass}`}>
+                {props.listItems.map((row, index) => {
+                    return <DynamicListItem key={`dynamic-list-row-${index}`} {...row} />;
+                })}
+            </ul>
+        );
+    } else {
+        list = <div role="alert">{props.noResultsText}</div>;
```
rav-pradhan, line 27: It'll fundamentally depend on how urgent we think it is for the user to be notified that no results have appeared, but my initial feeling is that it might be preferable to use `aria-live="polite"` instead, as the I get the impression this wouldn't be a time-critical notification for the user to hear

src/app/components/dynamic-list/DynamicListItem.jsx
```diff
@@ -0,0 +1,55 @@
...
+    buttonName: PropTypes.string,
+    buttonCallback: PropTypes.func,
+};
+
+const DynamicListItem = props => {
+    const colorBlack = "#000000";
+    let icon;
+    switch (props.icon) {
+        case "Shield-person":
+            icon = <PersonInShield classes="dynamic-list-item__icon" viewBox="0 0 15 15" fillColor={colorBlack} ariaLabel={props.iconName} />;
```
rav-pradhan, line 22: Maybe consider extracting the `viewBox` property into its own variable, similar to what you've done with `colorBlack`. On the off chance we ever need to update it, we could do so in one place rather than three

src/app/components/dynamic-list/DynamicList.jsx
```diff
@@ -0,0 +1,46 @@
...
+    }
+    return (
+        <div>
+            <span className="dynamic-list__title">{props.title}</span>
```
rav-pradhan, line 31: If this is a title, should we use a suitable heading element instead of `span`? May require additional CSS to override any global heading tag styles, but it feels like it would be more semantically accurate

src/app/components/dynamic-list/DynamicListItem.jsx
```diff
@@ -0,0 +1,55 @@
...
+            icon = <TickInShield classes="dynamic-list-item__icon" viewBox="0 0 15 15" fillColor={colorBlack} ariaLabel={props.iconName} />;
+            break;
+        case "Person":
+            icon = <Person classes="dynamic-list-item__icon" viewBox="0 0 15 15" fillColor={colorBlack} ariaLabel={props.iconName} />;
+            break;
+    }
+    return (
+        <li className="dynamic-list-item">
+            {icon}
+            <span className="dynamic-list-item__title">{props.title}</span>
```
rav-pradhan, line 34: Similar to in the `DynamicList` component, should we be using a heading element here instead of a `span`?

rav-pradhan (commented):
I've only gone through the code, but I think it's looking good and I've only left a few fairly minor comments. Let me know if you have any thoughts with them

src/scss/components/_dynamic_list.scss
dandwal: I have no idea. 
This is an awful variable name but seems to be used throughout Florence not something I have added. but a lot of padding margins and even text heights have been set with it.

src/scss/components/_dynamic_list.scss
dandwal: Yes, and no. 21px is on the UX design. It is used for titles around other areas in Florence but often declared like `$magicnumber1 * 3` which I could change this to if preferred?

src/app/components/dynamic-list/DynamicList.jsx
dandwal: So good spot around the `aria-live` should be using that here rather than role as this is more of a live element. However, would argue this should be assertive and not polite.

The visual user journey is they use the 'search' bar to filter results. As they type they get instant feedback on results. If there are no results mid typing they know to stop typing as typing more characters won't yield a better result. Likewise they get visual feedback as to what character was the last one they typed that meant their were no results which can help indicate a mispelling, and they will know where that mispelling took place (last character they typed).

Now taking a look at the visually impaired user that relies on audio. If see to polite it will wait for the user to go idle before giving feedback. Meaning they will have to type very slowly one character at a time to get any feedback. A user who opts to use the filter will be using it with an idea of who they are looking for so are likely to type fast. e.g. the word 'Jane'. If we had no Janes but a Jared when they type 'n' there would be no results but they would have typed an extra 'e' meaning polite would have been triggered but never gone idle as the user types an 'e' next so they never hear it (with VoiceOver at least this means nothing is announced when on polite, unless typing very very slowly). With assertive as soon as there are no results they will get this instant feedback just like a visual user.

Happy to have further feedback on this as not done an accessible instant results field before. Another option is to delay the filter - but feel that would be a worse user journey for every user type.

src/app/components/dynamic-list/DynamicListItem.jsx
dandwal: I think in this instance the title isn't a heading per se, might be that 'title' isn't an accurate word to use here (not sure what would be a better fit for 'main list item text field'. 

I think this item is styled similar to a heading; bigger and bolder. But it doesn't semantically section content off in the hierarchy of page content. It can be that there is just a list of these headings but no content that follows each of them.

In addition, this component could be used as a subsection of a multitude of heading levels (depending on where on the page it is used, could be nested under a h3 meaning it should be an h4 (to keep the order of headings correct). So each time it is used we would have to pass in a prop and the parent would have to declare what level it should be.

I also think by doing this screen readers may get confused thinking there are 2000+ section on this page where really this is just a list of results (we currently have 1800 Florence users which will be in this list).

src/scss/components/_dynamic_list.scss
rav-pradhan: Fair, may be worth changing the name to base or something, but I won't hold you to it

src/app/components/dynamic-list/DynamicList.jsx
rav-pradhan: I think that's fair enough - thanks for the reasoning

src/app/components/dynamic-list/DynamicListItem.jsx
rav-pradhan: Yeah, could be quite messy with that many users. Think it's fine to leave as is for now then as it's not so much a title as you say

rav-pradhan (approved):
Happy with the changes, and clarifying comments make sense 👍🏽

PR #639, opened 2021-12-01
bug:  Edit collection

AnnaEmson:
### What

This is missing part merged before in https://github.com/ONSdigital/florence/pull/634 
The validation should only happen when there is actually a change in name, and should not stop saving other changes 

### How to review

### Who can review

Front end


rafahop (approved):
Looks good to me 👍

PR #640, opened 2021-12-01
Refactor/class components to func josh dataset upload details

JoshSinghONS:
### What

Updated src/app/views/uploads/dataset/upload-details/DatasetUploadDetails.jsx
Ignore src/app/views/logs/Logs.jsx is part of another PR that this PR accidentally inherited when branching.

### How to review

Florence command is make debug ENABLE_DATASET_IMPORT=true ENCRYPTION_DISABLED=true
Page required to navigate to is Florence Homepage -> Datasets in Navbar -> Click CPIH hyperlink

What happens

Screenshot 2021-12-01 at 14 08 55

What should happen

Screenshot 2021-12-01 at 14 09 23

Extra Info

Originally the console errors were telling me that repeatUploadStatusCheck() was undefined. However I did something which has caused that error to be replaced by the less helpful errors in the first image above. 

What I did I have no idea but it has something to do with make clean / make dev / make debug ENABLE_DATASET_IMPORT=true ENCRYPTION_DISABLED=true / npm run dev

### Who can review

Anyone.


src/app/views/uploads/dataset/upload-details/DatasetUploadDetails.jsx
```diff
@@ -632,31 +580,48 @@ class DatasetUploadController extends Component {
                 );
             }
         }
-    }
+    };
 
-    render() {
```
dandwal, line 637: This won't be called. Remove this line and the ending `}` sp that the return statement isn't nested.

As this is a functional component, it expects a `return` statement so that when it is called (used) it returns something to render so we just need to make sure that all functional components return something

src/app/views/uploads/dataset/upload-details/DatasetUploadDetails.jsx
```diff
@@ -12,141 +12,98 @@ import http from "../../../../utilities/http";
...
-            () => {
-                this.getUploadStatus();
-            }
-        );
-        this.intervalID = setInterval(async () => {
-            if (!this.state.isFetchingDataset) {
-                this.setState({ isFetchingDataset: true }, () => {
-                    this.getUploadStatus();
-                });
+        intervalID = setInterval(async () => {
```
dandwal, line 68: IntervalID is declared above as a const either this is a new variable in which case we should explicitly state it or the const should be a `let` on line 21

PR #641, opened 2021-12-03
New create team screen

dandwal:
### What

New create a team screen for the new Auth system for creating new teams, note this no longer creates teams in Zebedee but instead in AWS Cognito. 

#### What this looks like
Note: Happy to give a demo
If attempting to run this yourself you will need to run the dp-identity-api then in the handlers in the dp-identity-api remove 'auth' checks (this checks against the permissions-api normally)
Finally, in this code in the `florence/src/index.js` file replace

```
<Route path={`${rootPath}/groups/create`} component={config.enableNewSignIn ? userIsAuthenticated(userIsAdmin(CreateTeam)) : null} />
```

With

```
<Route path={`${rootPath}/groups/create`} component={config.enableNewSignIn ? CreateTeam : null} />

```

Then run `make node-modules; make debug ENABLE_DATASET_IMPORT=true ENCRYPTION_DISABLED=true ENABLE_NEW_SIGN_IN=true` in the root of Florence

(Note nav bar looks different when signed in properly, but for development purposes for ease of development the route in the index file was altered temporarily to allow any user to go to this page without signing in)

##### Filter users

https://user-images.githubusercontent.com/47502916/145177647-d1892fe8-98d8-40e8-9e1a-5e7f3769dcac.mov

##### Change the team name, and exit when there are unsaved changes

https://user-images.githubusercontent.com/47502916/145177685-2b82df65-110f-40d5-9044-7a9338432dc2.mov

##### Adding and removing members from the team 

https://user-images.githubusercontent.com/47502916/145177817-8d9b079f-a20a-4a63-9264-7058b4a73624.mov

##### Create the team (note the screen it goes to hasn't been constructed yet but you can see the URL does go to a new page)

https://user-images.githubusercontent.com/47502916/145178049-a94ab6be-ed59-4636-a092-b3ac78106344.mov

Note this has been accessibility tested manually and via automation (google lighthouse) score can be seen below:

Screenshot 2021-12-08 at 08 57 29

### How to review

Check the code for best practice
Help identity any potential issues

### Who can review

Anyone except me


src/app/config/thunks.js
```diff
@@ -120,3 +118,145 @@ export const createCollectionRequest = collection => (dispatch, getState) => {
...
+                };
+                notifications.add(notification);
+            }
+            console.error(error);
+        });
+};
+
+export const addMembersToTeam = groupName => (dispatch, getState) => {
+    let promises = [];
+    const { newTeam } = getState().state?.newTeam ?? {};
```
AnnaEmson, line 226: it is better to avoid reaching to state in thunks https://blog.isquaredsoftware.com/2017/01/idiomatic-redux-thoughts-on-thunks-sagas-abstraction-and-reusability/

src/app/config/newTeam/newTeamReducer.js
```diff
@@ -0,0 +1,51 @@
...
+        case NEW_TEAM_ADD_USER: {
+            return {
+                ...state,
+                usersNotInTeam: state.usersNotInTeam.filter(filteredUser => filteredUser.email !== action.user.email),
+                usersInTeam: [...state.usersInTeam, action.user],
+            };
+        }
+        case NEW_TEAM_RESET: {
+            return {
+                ...initialState,
```
AnnaEmson, line 41: can you please do not mutate state in this way. I really do not understand what you are trying to achieve by doing this. you remove the whole state and there next line is to fetch it again? 
    const { dispatch, router, route, newTeam } = props;
    const [userConfirmedToLeave, setUserConfirmedToLeave] = useState(false);
    const [teamName, setTeamName] = useState("");
    useEffect(() => {
        dispatch(resetNewTeam());
        dispatch(getUsersRequest());
    }, []);

src/app/config/newTeam/newTeamReducer.js
```diff
@@ -0,0 +1,51 @@
...
+            return {
+                ...state,
+                usersInTeam: state.usersInTeam.filter(filteredUser => filteredUser.email !== action.user.email),
+                usersNotInTeam: [...state.usersNotInTeam, action.user],
+            };
+        }
+        case NEW_TEAM_ADD_USER: {
+            return {
+                ...state,
+                usersNotInTeam: state.usersNotInTeam.filter(filteredUser => filteredUser.email !== action.user.email),
```
AnnaEmson, line 35: those values should be done in selectors not in reducer. I don't think other places in app uses this value
usersNotInTeam and usersInTeam should be derived state

src/app/components/users/UsersNotInTeam.jsx
```diff
@@ -0,0 +1,70 @@
...
+        <div className="grid__col-4 margin-top--3">
+            <DynamicList
+                title="Add member to team"
+                headingLevel="2"
+                listItems={usersNotInTeamList}
+                noResultsText="No viewers found"
+                filterPlaceholder="Search viewers by name or email"
+                handleSearchInput={handleSearch}
+                listHeightClass="height--60"
+                loading={props.loading}
```
AnnaEmson, line 56: where this value is coming form, can't see it listed in proptypes on her  https://github.com/ONSdigital/florence/pull/641/files#diff-a1f8a8979809bd04cbca2731d05da0cdb90473bbfcdc64d5f2f89a774b39ed46R7

src/app/views/teams/team-create/CreateTeam.jsx
```diff
@@ -0,0 +1,175 @@
...
+                    &#9664; <Link to={url.resolve("../")}>Back</Link>
+                </span>
+                <h1 className="margin-top--1 margin-bottom--1">Create a preview team</h1>
+                {teamNameInputArea}
+                <span>
+                    <strong>Members</strong>
+                </span>
+                {newTeam.usersInTeam?.length > 0 ? teamsMemberChips : noTeamMembers}
+            </div>
+            <UsersNotInTeam loading={true} />
```
AnnaEmson, line 161: whys this is loading true always, if its not managed by parent certainly do not need to be passed ?

src/app/views/teams/team-create/CreateTeam.test.js
```diff
@@ -0,0 +1,85 @@
...
+            setRouteLeaveHook: () => {},
+        },
+    };
+    describe("Given valid props on initial load", () => {
+        const component = mount(
+            <WrapperComponent>
+                <CreateTeam {...defaultBaseProps} />
+            </WrapperComponent>
+        );
+        it("renders the relevant components to screen", () => {
```
AnnaEmson, line 69: this test seems to be exactly the same what snapshot does?

src/app/views/teams/team-create/CreateTeam.test.js
```diff
@@ -0,0 +1,85 @@
...
+            groups: [],
+            status: "CONFIRMED",
+            active: true,
+            id: "ggg-hhh-iii",
+            status_notes: "",
+        },
+    ],
+};
+
+jest.mock("../../../utilities/api-clients/user", () => {
```
AnnaEmson, line 48: don't see you use it?

src/app/views/teams/team-create/TeamCreate.jsx
```diff
@@ -9,6 +9,7 @@ const propTypes = {
     onCreateSuccess: PropTypes.func.isRequired,
 };
 
+// TODO remove when new AUTH is turned on
 class TeamCreate extends Component {
```
AnnaEmson, line 13: does it need to be a class component?

src/index.js
```diff
@@ -72,6 +73,16 @@ const userIsAdminOrEditor = connectedReduxRedirect({
     allowRedirectBack: false
 });
 
+const userIsAdmin = connectedReduxRedirect({
```
AnnaEmson, line 76: can't see it is used in your code, you used `userIsAdminOrEditor`

src/app/views/teams/team-create/TeamCreate.jsx
dandwal: This whole 'TeamCreate.jsx' file will be deleted, the new screens replace this entirely, this TODO that has been added is to make clean up later easier to ensure we delete the then unused components

src/app/config/thunks.js
AnnaEmson: `Dan Abramov - "Don't Use getState in thunks"
First, Dan Abramov wrote a Stack Overflow answer back in February, in which he said:

In general accessing state in action creators is an anti-pattern and you should avoid this when possible. The few use cases where I think it’s acceptable is for checking cached data before you make a request, or for checking whether you are authenticated (in other words, doing a conditional dispatch). Passing data such as state.something.items in an action creator is definitely an anti-pattern and is discouraged because it obscured the change history: if there is a bug and items are incorrect, it is hard to trace where those incorrect values come from because they are already part of the action, rather than directly computed by a reducer in response to an action. So do this with care.`

src/app/config/thunks.js
```diff
@@ -120,3 +118,145 @@ export const createCollectionRequest = collection => (dispatch, getState) => {
             console.error(error);
         });
 };
+
+export const getUsersRequest = () => dispatch => {
+    users
+        .getAll({ active: true })
+        .then(response => {
+            dispatch(getUsersRequestSuccess(response));
+        })
```
IncredibleMrTim, line 127: we can use async/await here...

```javascript
export const getUsersRequest = async () => dispatch => {
try {
      const result = await users.getAll({ active: true })
      if(result) dispatch(getUsersRequestSuccess(result));
} catch(e){
...
}        
```

src/app/config/thunks.js
```diff
@@ -120,3 +118,145 @@ export const createCollectionRequest = collection => (dispatch, getState) => {
...
+        .catch(error => {
```
IncredibleMrTim, line 128: the try/catch here would be better at source in `getAll()` in `src/app/utilities/api-clients/user.js`

src/app/config/thunks.js
```diff
@@ -120,3 +118,145 @@ export const createCollectionRequest = collection => (dispatch, getState) => {
...
+            // TODO move, handling errors should be done elsewhere see note in createCollectionRequest func above
+            if (error.status != null) {
```
IncredibleMrTim, line 130: this can be just `if(error.status) {...`

src/app/config/thunks.js
```diff
@@ -120,3 +118,145 @@ export const createCollectionRequest = collection => (dispatch, getState) => {
...
+                }
+            } else {
+                const notification = {
+                    type: "warning",
+                    message: errCodes.GET_USERS_UNEXPECTED_ERROR_SHORT,
+                    isDismissable: true,
+                };
+                notifications.add(notification);
+            }
+            console.error(error);
```
IncredibleMrTim, line 177: shouldn't log to the console in completed code

src/index.js
dandwal: Cheers, yes this was added and used then I removed the use of it and forgot to get rid of the function. But, since making this PR seems the permissions-api only allows Admins now to create teams so will be switching it back to use this function.

src/app/components/users/UsersNotInTeam.test.js
```diff
@@ -0,0 +1,36 @@
...
+        loading: false,
+    };
+    describe("Given valid props on initial load", () => {
+        const component = mount(
+            <WrapperComponent>
+                <UsersNotInTeam {...UsersNotInTeamProps} />
+            </WrapperComponent>
+        );
+        it("renders the relevant components to screen", () => {
+            expect(component.find("h2.dynamic-list__title")).toHaveLength(1);
```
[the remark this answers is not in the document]
AnnaEmson: Just to let you know. the testing library is available in the repo already. you can start using it as pleased on each test basis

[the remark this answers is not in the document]
AnnaEmson: here is one example I made https://github.com/ONSdigital/florence/blob/develop/src/app/components/selectable-box/double-column/DoubleSelectableBox.test.js

src/app/components/navbar/NavBar.jsx
```diff
@@ -83,12 +82,20 @@ const NavBar = props => {
                                 Users and access
                             </Link>
                         </li>
-
-                        <li className="global-nav__item">
-                            <Link to={`${rootPath}/teams`} activeClassName="selected" className="global-nav__link">
-                                Teams
-                            </Link>
-                        </li>
+                        {props.config?.enableNewSignIn && (
```
AnnaEmson, line 85: would be good to add this part to NavBar.jsx test ?

src/app/config/newTeam/newTeamReducer.js
AnnaEmson: I don't think those objects are being used outside this functionality and any other component is subscribed to those changes, where you would need to use props drilling.
in situations like this you just need to calculate two separate lists based on one state in store `allUsers` , as at this is the only source of true about who belongs where.

src/app/config/thunks.js
AnnaEmson: understand your pain, but  once it is done properly, where there is a clear separation of concerns  those components are easy to maintain and test. React is very easy and beautiful.  At the moment there is a mess in Florence  and lots of different sources of data and this is not maintainable in long run.

src/app/components/navbar/NavBar.test.js
```diff
@@ -63,6 +64,23 @@ describe("NavBar", () => {
...
+            const props = {
+                ...defaultProps,
+                config: {
+                    ...defaultProps.config,
+                    enableNewSignIn: true,
+                },
+            };
+            const component = shallow(<NavBar {...props} user={authenticatedUser} />);
+            const nav = component.find(Link);
+            it("third item should be named 'Preview teams'", () => {
```
AnnaEmson, line 77: do we need to be so specific it needs to be third? I don't see the assertion for this in the test anyway? you checking the class and loops thru its elements checking if they have a child of `NavbarItemsWithCognitoAuth` ?  I can't see where you checking if navigation has `Preview teams` item. Please make sure test are simple. 
Loops in test should be avoided if possible.

src/app/components/users/UsersNotInTeam.test.js
```diff
@@ -0,0 +1,36 @@
...
+        dispatch: mockDispatch,
+        loading: false,
+    };
+    describe("Given valid props on initial load", () => {
+        const component = mount(
+            <WrapperComponent>
+                <UsersNotInTeam {...UsersNotInTeamProps} />
+            </WrapperComponent>
+        );
+        it("renders the relevant components to screen", () => {
```
AnnaEmson, line 23: can you please assert the props you passed to the component in line 20 . Also I can't see the test for other props passed to the component: 
`addUserToTeam: PropTypes.func,
    usersNotInTeam: PropTypes.arrayOf(PropTypes.object),
    allPreviewUsers: PropTypes.arrayOf(PropTypes.object),` 
    there is so much logic in this simple component it needs to be tested

src/app/config/newTeam/newTeamReducer.js
```diff
@@ -0,0 +1,29 @@
+import { GET_USERS_REQUEST_SUCCESS, NEW_TEAM_UNSAVED_CHANGES } from "../constants";
+
+export const initialState = {
+    allPreviewUsers: [],
+    unsavedChanges: false,
+};
+
+const newTeamReducer = (state = initialState, action) => {
```
AnnaEmson, line 8: This reducer hold seta of users not teams.the naming is confusing. Im te action you fetched users so it is the same as UsersReducer and should hold state of 
users: 
loadingUsers
etc... 
I am not sure what is the purpose for unsavedChanges ?

AnnaEmson: can you use the current user state ?

src/app/config/thunks.js
AnnaEmson: I agree with Dan, no harm in having this here

src/app/config/selectors.js
```diff
@@ -18,8 +18,12 @@ export const getFilteredCollections = createSelector(getMappedCollections, getSe
         return name.includes(string);
     });
 });
+
 export const getWorkingOn = state => state.global.workingOn;
 export const getActive = state => state.collections.active;
 export const getCollectionsLoading = state => state.collections.isLoading;
 export const getCollectionCreating = state => state.collections.isCreating;
 export const getIsUpdatingCollection = state => state.collections.isUpdating;
+export const selectNewTeamUsersNotInTeam = state => state.newTeam.usersNotInTeam;
```
AnnaEmson, line 27: is this needed.. again naming confusing. Worth checking  https://redux.js.org/style-guide/style-guide#organize-state-structure-based-on-data-types-not-components

src/app/config/store.js
```diff
@@ -16,6 +17,7 @@ export const store = createStore(
     combineReducers({
         state: reducer,
         user: userReducer,
+        newTeam: newTeamReducer,
```
AnnaEmson, line 20: `teams` ?

AnnaEmson: but still I am not sure if you need creating new reducer at all.

src/app/config/thunks.js
```diff
@@ -145,3 +148,95 @@ export const approveCollectionRequest = (id, redirect) => dispatch => {
             collectionDetailsErrorNotifications.updateCollection(error);
         });
 };
+
+export const getUsersRequest = () => dispatch => {
+    users
+        .getAll({ active: true })
+        .then(response => {
```
AnnaEmson, line 155: what if the response fails ?

src/app/views/teams/team-create/CreateTeam.test.js
```diff
@@ -0,0 +1,78 @@
...
+            status: "CONFIRMED",
+            active: true,
+            id: "ggg-hhh-iii",
+            status_notes: "",
+        },
+    ],
+};
+
+describe("CreateTeam", () => {
+    const defaultBaseProps = {
```
AnnaEmson, line 49: `props` ?

src/app/views/teams/team-create/CreateTeam.test.js
AnnaEmson: it would be nice to actually test the functionality here in this component. the name suggests `create team` can you add tests for events?

src/app/components/dynamic-list/DynamicList.jsx
```diff
@@ -25,12 +26,16 @@ const DynamicList = props => {
...
         );
+    } else if (props.loading) {
+        list = <div className="loader loader--dark" />;
     } else {
         list = <div aria-live="assertive">{props.noResultsText}</div>;
     }
     return (
         <div>
-            <HeadingLevel className="dynamic-list__title">{props.title}</HeadingLevel>
+            <HeadingLevel className="dynamic-list__title" data-testid="dynamic-list-title">
```
AnnaEmson, line 36: are you using this `data-testid="dynamic-list-title"` ? I haven't seen test for this component, did you forget to push it?

src/app/components/dynamic-list/__snapshots__/DynamicList.test.jsx.snap
```diff
@@ -4,6 +4,7 @@ exports[`DynamicList given props does it create a list matches the snapshot 1`]
 <div>
   <h1
     className="dynamic-list__title"
+    data-testid="dynamic-list-title"
```
AnnaEmson, line 7: as above, you do not query this element by this id in any of your test at the moment. might by leftover or you forgot to include the file?

src/app/components/navbar/NavBar.jsx
```diff
@@ -83,12 +82,20 @@ const NavBar = props => {
...
+                            <li className="global-nav__item">
+                                <Link to={`${rootPath}/groups`} activeClassName="selected" className="global-nav__link">
+                                    Preview teams
```
AnnaEmson, line 88: what is the difference between `Preview teams` and `Teams`? you made the routes to `/groups`. We need to make sure next dev come after you and understand what it this about I think using consistent naming would help

src/app/utilities/api-clients/teams.js
```diff
@@ -14,7 +14,19 @@ export default class teams {
...
+        });
+    }
+
+    static createTeam(body) {
+        return http.post(`/groups`, body).then(response => {
+            return response;
+        });
+    }
+
+    static addMemberToTeam(teamID, userID) {
```
AnnaEmson, line 28: this should be consistent you use  `addMemberToTeam` but the api is groups...

AnnaEmson:
Hi Daniel, I tried to make a review of this PR. First you need to update branch, second I am really struggling here, would you be able to spare me 15 min and go over it with me please

src/app/utilities/api-clients/teams.js
dandwal: Hi, yer so we are using AWS Cognito 'groups' for storing 'teams' and 'roles' (the new backend for Auth rather than Zebedee)

A user can be in a Cognito 'group' for a role e.g. admin, publisher (some others will exist) which grant access to different areas of Florence and content.
A user can also be in a group for a team (which when attached to a collection, gives the team access to content).

So where in Zebedee we had Roles and Teams as completely different things we now just have 'groups' and are using that for both roles and teams. (as you attach policies to groups that grant levels of access).

Content Designer was also consulted in regards to terminology around showing the user 'teams' or 'groups' and it was decided by them that team was more appropriate.

So we are calling off to add members to a team, but underneath the hood, teams are stored as 'groups'. I thought a good place to make that distinction in the code is when dealing with the API (so in the API client). So to Florence, everything is 'team' or 'role' like what is presented to users on the frontend, but the backed / APIs it is 'groups'.

src/app/components/navbar/NavBar.jsx
dandwal: 'Team' is a legacy term users used to see
'Preview team' is the correct term going forward for displaying to the user

As there will only be one type of 'team' in Florence which is a preview 'team' I think it is safe to refer to 'preview team' as just 'team' in the code as we have been doing all along, the only real change is that users see a more descriptive name of what a 'team' is.

A Preview team is a collection of users who are not publishers nor admins (so Business area users rather than Publishing team users in DP). These users get added to a Preview Team and that Preview Team is added to collections within Florence and those collections contain content due to be published. This gives the Business areas the opportunity to read publications and review them before they get published.

The term itself was decided upon by our Content Designer after they consulted with the Publishing team. So going forward the correct full term for what was 'teams' is now 'Preview Teams' (original Florence I don't think had any content designer involved). (note it is actually technically possible to add Publishers and Admins to preview teams, but this grants them no extra access as they will already have access due to their role).

Hope that clears that up a little, in terms of 'groups' I have added a comment onto another thread you commented on in the api-clients.

src/app/components/dynamic-list/DynamicList.jsx
dandwal: Ah yes, the test uses the className 'dynamic-list__title' but should instead use this new ID

src/app/config/thunks.js
dandwal: As per Tims PR comment handling of this has been moved to the file `utilities/api-clients/user.js` it will call off to fire a notification (depending on the error) e.g. "An unexpected error has occurred whilst trying to get users. Please try again shortly, if the error persists please contact an administrator." Then finally in here it will log the error to console

src/app/config/store.js
dandwal: Happy to change the name on this as I am not too happy with 'newTeam' either but couldn't think of a more appropriate name at the time. Since PRing this it has changed a lot, as it used to contain `usersInTeam` `usersNotInTeam` and `allPreviewUsers`, but now it is just `allPreviewUsers` and `unsavedChanges`. So this could contain the upcoming 'edit' team data and list of teams data too.

Aware there is state.teams but this is cumbersome and that datastore is more tailored to what Zebedee returns rather than the new system Cognito for user and team management. - Will change the name to 'teams'. Although allPreviewUsers may not fit so nicely as they are not 'in' the team yet - but `allPreviewUsers` is used after being populated to list all possible users an admin could choose to add to a new team, then that data field is used to keep track of users that are in a basket ready to be added to the new team, as if they aren't in that field then they are in a basket to be added to the team.

src/app/config/selectors.js
dandwal: Have removed the one highlighted here (as it is unused) and renamed the two under it to be more descriptive of what they are storing (a little verbose but better to be clear and verbose than short and unclear)
`selectNewTeamAllPreviewUsers` -> `selectPreviewUsersNotInCurrentTeam` this is the list of preview users not assigned to the current team. These are the equivalent to users that used to have the 'role' `viewer` in Zebedee and who are not in the current team the user is creating (but could be in other teams)
`selectNewTeamUnsavedChanges` -> `selectUnsavedChangesOnTeamCreation`

src/app/config/newTeam/newTeamReducer.js
dandwal: Although it is storing 'users' the context of these users are users in a basket ready to be added to a new team (that has yet to be created). 

unsavedChanges is used to store whether or not the user has unsaved changes when editing(editing to come) or creating a new team. E.g. if they enter data into the 'team name' field but then click back a popup will appear checking if they want to leave the page or stay on the page as they have unsaved changes. In addition this is used in the bottom bar banner which has the option to save or cancel, if the user has unsaved changes then an exclamation mark in a triangle appears along with the text 'There are unsaved changes' - this is as per UX designs

AnnaEmson: in that situation think `unsavedChanges` should be a local state not global state in store as no other component will have use of this

src/app/config/initialState.js
```diff
@@ -38,4 +40,5 @@ export const initialState = {
     preview: {
         selectedPage: null,
     },
+    Teams: teamsReducerInitialState,
```
AnnaEmson, line 43: `teams`?

src/app/components/dynamic-list/DynamicList.test.jsx
```diff
@@ -34,7 +34,7 @@ describe("DynamicList", () => {
         };
         const component = mount(<DynamicList {...dynamicListProps} />);
         it("should display a dynamic list", () => {
-            expect(component.find(".dynamic-list__title").length).toBe(1);
+            expect(component.find('[data-testid="dynamic-list-title"]').exists()).toBe(true);
```
AnnaEmson, line 37: the `data-testid` is the api for react testing library which you are not using here, you use enzyme in this file. 
the way you use it in test is :
`ByTestId('dynamic-list-title') ` for example `const element = screen.getByTestId('dynamic-list-titlet')`
sorry Tim confused you here I think, you should revert to classname :-)

src/app/components/navbar/NavBar.test.js
```diff
@@ -63,6 +63,22 @@ describe("NavBar", () => {
...
+                config: {
+                    ...defaultProps.config,
+                    enableNewSignIn: true,
+                },
+            };
+            const component = shallow(<NavBar {...props} user={authenticatedUser} />);
+            it("'Preview teams' option should be present", () => {
+                expect(component.hasClass("global-nav__list")).toBe(true);
+                expect(component.find(Link)).toHaveLength(NavbarItems.length);
+                expect(component.find('[data-testid="preview-teams"]').getElement().props.children[0].includes("Preview teams"));
```
AnnaEmson, line 78: the same comment as above about `data-testid`

dandwal:
New GPG key, old one is removed so CI won't like these commit signatures. Replaced with https://github.com/ONSdigital/florence/pull/676

PR #644, opened 2021-12-06
Feature/search component testing

AnnaEmson:
### What

Add test for Search component
Add mock data to share across components
Add wrapper component for testing 
Organise and correct some other test components 
correct typo in Collections propTypes 

### How to review

### Who can review

### DIFF
--- a/src/app/components/search/Search.test.js
+++ b/src/app/components/search/Search.test.js
@@ -1,0 +1,28 @@
+import React from "react";
+import { WrapperComponent } from "../../../tests/test-utils";
+import Search from "./Search.jsx";
+import { mount, shallow } from "enzyme";
+import renderer from "react-test-renderer";
+
+describe("Search", () => {
+    it("matches the snapshot", () => {
+        const wrapper = renderer.create(
+            <WrapperComponent>
+                <Search />
+            </WrapperComponent>
+        );
+        expect(wrapper.toJSON()).toMatchSnapshot();
+    });
+
+    it("updates value on change", () => {
+        const wrapper = mount(
+            <WrapperComponent>
+                <Search />
+            </WrapperComponent>
+        );
+        const event = { target: { value: "test" } };
+        const onChange = jest.fn();
+        wrapper.find("input#search_input").simulate("change", event);
+        expect(wrapper.find("input#search_input").getElement().props.value).toBe("test");
+    });
+});
--- a/src/app/components/selectable-box/SelectableBoxController.test.js
+++ b/src/app/components/selectable-box/SelectableBoxController.test.js
@@ -1,5 +1,5 @@
 import React from "react";
-import SelectableBoxController from "./SelectableBoxController.jsx";
+import SelectableBoxController from "./SelectableBoxController";
 import renderer from "react-test-renderer";
 import { shallow, mount } from "enzyme";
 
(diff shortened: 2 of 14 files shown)


src/app/components/selectable-box/SelectableBoxController.test.js
```diff
@@ -1,7 +1,8 @@
 import React from "react";
-import SelectableBoxController from "./SelectableBoxController.jsx";
+import SelectableBoxController from "./SelectableBoxController";
 import renderer from "react-test-renderer";
 import { shallow, mount } from "enzyme";
+import JestSnapshot from "../../../node_modules/jest-snapshot/build/index";
```
dandwal, line 5: More of a question than anything, Does this need to be imported here for anything?
I can see it was using snapshots before without this import.

src/app/components/selectable-box/double-column/DoubleSelectableBoxController.test.js
```diff
@@ -0,0 +1,44 @@
+import React from "react";
+import DoubleSelectableBoxController from "./DoubleSelectableBoxController";
+import renderer from "react-test-renderer";
+import { shallow, mount } from "enzyme";
```
dandwal, line 4: Don't think shallow is used in this file, no need to import it

src/tests/mockData.js
```diff
@@ -0,0 +1,95 @@
...
+      collectionOwner: "PUBLISHING_SUPPORT",
+      timeseriesImportFiles: [],
+      id: "test-sau39393uyqha8aw8y3n3",
+      name: "Complete collection",
+      type: "manual",
+      teams: ["Team 2"],
+  },
+];
+
+const items = [
```
dandwal, line 59: Wonder if we should split this file out a bit maybe things like
- `/tests/mock/collections`
- `/tests/mock/users` 
etc 

I worry people will just add more and more to this file with just slight adjustments then after a while people will just get into the habit of adding things to the file without checking what is there as it becomes rather massive and we could see duplicates.

Also, wonder if we should change this name `items` to something more meaningful, I had to see how it was used to work out what it was for. Might just be `doubleSelectableBoxControllerItems` or something.

dandwal (commented):
LGTM minor comments

src/app/components/selectable-box/SelectableBoxController.test.js
AnnaEmson: oh that is bad.. it was added by auto typing by my VSC, bad boy ... I missed it

src/app/components/search/Search.test.js
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import { WrapperComponent } from "../../../tests/test-utils";
+import Search from "./Search.jsx";
+import { mount, shallow } from "enzyme";
+import renderer from "react-test-renderer";
+
+describe("Search", () => {
+    xit("matches the snapshot", () => {
```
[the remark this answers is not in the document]
AnnaEmson: missed it, sorry

src/tests/mockData.js
AnnaEmson: I would leave this fir now as it is. I named the collections as per props name they are requested

dandwal (approved):

PR #645, opened 2021-12-07
Fix/adding collection id into cookie

mr-nick17:
### What

Added collection id into cookie when user selects collection

### How to review

Sense check
Tests pass

### Who can review

Anyone but me

### DIFF
--- a/src/app/views/collections/Collections.jsx
+++ b/src/app/views/collections/Collections.jsx
@@ -26,6 +26,7 @@ const Collections = props => {
             props.dispatch(push(`${props.rootPath}/collections/${id}/preview`));
             return;
         }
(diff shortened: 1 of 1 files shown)


AnnaEmson (approved):

src/app/views/collections/Collections.jsx
```diff
@@ -26,6 +26,7 @@ const Collections = props => {
             props.dispatch(push(`${props.rootPath}/collections/${id}/preview`));
             return;
         }
+        cookies.add("collection", id, null);
```
dandwal, line 29: Should we just add this above line 23 and remove line 24? (so that we don't need to write the same line twice?

dandwal (approved):
LGTM

PR #650, opened 2021-12-15
bug: florence sort collection by name 

AnnaEmson:
### What

Fix the sort on name as it was using wrong default sort direction. 

### How to review
code meets standards and functionality is working 

### Who can review
all


src/app/components/selectable-box/double-column/DoubleSelectableBox.jsx
```diff
@@ -27,8 +27,8 @@ const DoubleSelectableBox = props => {
     const { sortedItems, requestSort, sortConfig } = useSort(props.items, activeSort);
 
     const handleSortClick = (key, direction) => {
-        setActiveSort({key, direction});
-        requestSort({key, direction});
+        setActiveSort({ key, direction });
+        requestSort({ key, direction });
```
rafahop, line 31: I wonder if we really need the `direction` param in `handleSortClick` as it seems to always be "ASC"?

AnnaEmson: yes, I would like to leave it as it uses the useSort function and and would like it to be generic for reusability. no harm in sending this value

src/app/hooks/useSort.js
```diff
@@ -32,13 +32,11 @@ export const useSort = (items, config = null) => {
...
-        let {key, direction} = activeSort
-        if (sortConfig && sortConfig.key === key && sortConfig.direction === "ASC") {
-            direction = "DESC";
-        }
-
-        setSortConfig({ key, direction });
+    const requestSort = key => {
+        setSortConfig(prevState => {
+            if (prevState.direction === "ASC") return { key, direction: "DESC" };
+            if (prevState.direction === "DESC") return { key, direction: "ASC" };
```
rafahop, line 38: Potentially we don't need this second if statement? That way we would also cover the case where we mistakenly pass a wrong direction value.

```suggestion
            return { key, direction: "ASC" };
```

src/app/components/selectable-box/double-column/DoubleSelectableBox.jsx
```diff
@@ -5,30 +5,10 @@ import { useSort } from "../../../hooks/useSort";
...
-
 const DoubleSelectableBox = props => {
-    const [activeSort, setActiveSort] = useState({ key: "name", direction: "ASC" });
-    const { sortedItems, requestSort, sortConfig } = useSort(props.items, activeSort);
-
-    const handleSortClick = (key, direction) => {
-        setActiveSort({key, direction});
-        requestSort({key, direction});
+    const { sortedItems, requestSort, sortConfig } = useSort(props.items, { key: "name", direction: "ASC" });
+    const handleSortClick = key => {
```
rafahop, line 10: Is this redundant now? Should we just call `requestSort(...)` instead?

AnnaEmson: good idea

src/app/hooks/useSort.js
AnnaEmson: nope this needs to stay as the direction changes and manages the arrows selections

rafahop (approved):

PR #653, opened 2021-12-22
fix: make a check if default is available

AnnaEmson:
### What

some linting changes and sanity checks on default values 

### How to review

sanity check

### Who can review

all devs


mr-nick17 (approved):
LGTM

PR #667, opened 2022-02-01
Feature/add flexible cantabular dataset type

mr-nick17:
### What

Reworked logic to handle all `formats` of type `cantabular`

### How to review

Sense check

### Who can review

Anyone but me


andre-urbani (approved):
LGTM

PR #676, opened 2022-02-14
Create team screen

dandwal:
### What

New create a team screen for the new Auth system for creating new teams, note this no longer creates teams in Zebedee but instead in AWS Cognito. 

#### What this looks like
Note: Happy to give a demo
If attempting to run this yourself you will need to run the dp-identity-api then in the handlers in the dp-identity-api remove 'auth' checks (this checks against the permissions-api normally)
Finally, in this code in the `florence/src/index.js` file replace

```
<Route path={`${rootPath}/groups/create`} component={config.enableNewSignIn ? userIsAuthenticated(userIsAdmin(CreateTeam)) : null} />
```

With

```
<Route path={`${rootPath}/groups/create`} component={config.enableNewSignIn ? CreateTeam : null} />

```

Then run `make node-modules; make debug ENABLE_DATASET_IMPORT=true ENCRYPTION_DISABLED=true ENABLE_NEW_SIGN_IN=true` in the root of Florence

(Note nav bar looks different when signed in properly, but for development purposes for ease of development the route in the index file was altered temporarily to allow any user to go to this page without signing in)

##### Filter users

https://user-images.githubusercontent.com/47502916/145177647-d1892fe8-98d8-40e8-9e1a-5e7f3769dcac.mov

##### Change the team name, and exit when there are unsaved changes

https://user-images.githubusercontent.com/47502916/145177685-2b82df65-110f-40d5-9044-7a9338432dc2.mov

##### Adding and removing members from the team 

https://user-images.githubusercontent.com/47502916/145177817-8d9b079f-a20a-4a63-9264-7058b4a73624.mov

##### Create the team (note the screen it goes to hasn't been constructed yet but you can see the URL does go to a new page)

https://user-images.githubusercontent.com/47502916/145178049-a94ab6be-ed59-4636-a092-b3ac78106344.mov

Note this has been accessibility tested manually and via automation (google lighthouse) score can be seen below:

Screenshot 2021-12-08 at 08 57 29

### How to review

Check the code for best practice
Help identify any potential issues

### Who can review

Anyone except me


src/app/config/reducer.js
```diff
@@ -1,5 +1,6 @@
 import { initialState } from "./initialState";
 import * as types from "./constants";
+import { GET_USERS_REQUEST_SUCCESS } from "./constants";
```
AnnaEmson, line 3: you don't need this

src/app/config/reducer.js
```diff
@@ -485,6 +486,15 @@ export default function reducer(state = initialState, action) {
                 isCreating: false,
             };
         }
+        case GET_USERS_REQUEST_SUCCESS: {
```
AnnaEmson, line 489:
```suggestion
        case types.GET_USERS_REQUEST_SUCCESS: {
```

AnnaEmson (approved):
still few things could be improved, but approving as don't want to stop the development as this is few months in PR already. we can refactor it later as tech debts

PR #678, opened 2022-02-16
Feature/update post body for Cantabular types

mr-nick17:
### What

- Updated POST /datasets body to
```
is_based_on: {
    id: this.state.isBasedOn,
    type: this.state.format,
},
type: this.state.format,
```
- Changed conditionals to look for 'includes' `cantabular` rather than being specific types
- Refactored `CreateCantabularDatasetController` to perform a GET instead of relying on query string parameters 
- Updated unit tests

### How to review

- Sense check
- Tests pass

### Who can review

Anyone but me

### DIFF
--- a/src/app/views/datasets-new/create/CreateCantabularDatasetController.jsx
+++ b/src/app/views/datasets-new/create/CreateCantabularDatasetController.jsx
@@ -3,6 +3,7 @@ import { push } from "react-router-redux";
 import PropTypes from "prop-types";
 
 import datasets from "../../../utilities/api-clients/datasets";
+import recipes from "../../../utilities/api-clients/recipes";
 import notifications from "../../../utilities/notifications";
 import url from "../../../utilities/url";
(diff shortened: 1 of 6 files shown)


src/app/views/datasets-new/create/CreateCantabularDatasetController.jsx
```diff
@@ -24,24 +25,57 @@ export class CreateCantabularDatasetController extends Component {
...
 
+    handleGETSuccess = recipe => {
+        this.setState({
+            format: recipe.format,
+            isBasedOn: recipe.cantabular_blob,
+        });
+    };
+
+    getRecipe = () => {
+        const recipeId = this.props.params.recipeID;
```
joegasewicz, line 54: Just a question, can `this.props.params` be `undefined` - if so, `this.props.params.recipeID` might throw a reference error....

joegasewicz (approved):
LGTM

PR #679, opened 2022-02-17
Fix/amending json object props to include @

mr-nick17:
### What

@ symbol is required on the child props of `is_based_on`

### How to review

Sense check
Tests pass

### Who can review

!me


andre-urbani (approved):
LGTM

PR #680, opened 2022-02-18
feature: 704 - add groups to collections - Feed Collection Users with…

AnnaEmson:
… groups fetched form new api

### What

Ticket: [#704: Update collection viewer access teams list to use the identity API](https://trello.com/c/yNWv9HWq/704-704-update-collection-viewer-access-teams-list-to-use-the-identity-api)

In this PR:
1. correct prop types error form last PR
2. add check on config to fetch groups form new api. 
3. transition teams to groups for unity with new api 
4. correct styling of global component Select as had 

As this is still not ready feature, I hide all new functionality behind feature flag. I am not able to tets the collection creating part so this will be revised later. 
Screenshot 2022-02-18 at 15 55 57
Screenshot 2022-02-18 at 15 56 21
Screenshot 2022-02-18 at 16 23 35

ezgif com-gif-maker (11)

### How to review

please reach out for demo
check the code if meets the standards

### Who can review

front end dev please


src/app/components/selected-items/SelectedItem.jsx
```diff
@@ -16,7 +16,7 @@ function SelectedItem({ id, name, removeClassNames, classNames, handleRemoveItem
 SelectedItem.propTypes = {
     id: PropTypes.string.isRequired,
     name: PropTypes.string.isRequired,
-    onRemoveItem: PropTypes.func.isRequired,
```
dandwal, line 19: Do the following places also need an update of of the prop `onRemoveItem` -> `handleRemoveItem` https://github.com/ONSdigital/florence/blob/9880d4843691fec453d4ad616244976e27c6c33a/src/app/views/collections/edit/CollectionEdit.jsx#L144

https://github.com/ONSdigital/florence/blob/11f5fb6fa1d61932f6ec9728bae40abfa6aedd3b/src/app/views/users/groups/User.jsx#L16

https://github.com/ONSdigital/florence/blob/c3a0476d1e27c85cbccfc265894d980bd6b50961/src/app/views/collections/create/CreateNewCollection.jsx#L320

src/app/config/selectors.js
```diff
@@ -26,18 +26,20 @@ export const getCollectionCreating = state => state.collections.isCreating;
...
-    return teams.map(team => ({ id: team.id.toString(), name: team.name }));
+export const getGroups = state => state.groups.all;
+export const getGroupsLoading = state => state.groups.isLoading;
+
+export const getEnableNewSignIn = state => state.config.enableNewSignIn;
+
+export const getMappedGroups = createSelector(getGroups, getEnableNewSignIn, (groups, isNewSignIn) => {
+    if (!groups) return [];
+    return isNewSignIn
+        ? groups.map(group => ({ id: group.group_name, name: group.group_name }))
```
dandwal, line 37: I think the `name` field should have its information derived from the `description` field.
the `group_name` we are using as an ID but the `description` field is what is being used to populate the 'user-friendly' name. group_name is an autogenerated UID whereas `description` is the name the user enters when creating the team. It is worth noting, API users can create teams with no `descriptions`, because of this we should fall back and display the `group_name` as the `name` when there is no `description` to populate `name` with.

src/app/config/thunks.js
```diff
@@ -143,53 +143,67 @@ export const approveCollectionRequest = (id, redirect) => dispatch => {
...
                 }
-                case "FETCH_ERR": {
-                    const notification = {
-                        type: "warning",
-                        message: "There's been a network error whilst trying to get teams. Try refresh the page.",
-                        isDismissable: true,
-                    };
-                    notifications.add(notification);
-                    break;
+                console.log(error);
```
dandwal, line 160:
```suggestion
                console.error(error);
```

src/app/config/thunks.js
```diff
@@ -143,53 +143,67 @@ export const approveCollectionRequest = (id, redirect) => dispatch => {
...
+            })
+        : teams
+            .getAll()
+            .then(response => {
+                if (!response) return dispatch(actions.loadGroupsFailure());
```
dandwal, line 165: Not sure one line `if` statements have been used elsewhere in Florence, I think it is worth adding the braces around the block section even if it is more verbose. Makes it more readable and helps people who may be refactoring/modifying this section in the future not to accidentally miss the if statement.

src/app/utilities/tests/mockData.js
```diff
@@ -181,28 +181,31 @@ const teams = [
...
-        name: "Test Team",
-        members: [],
-        path: "test_team_1",
+        creation_date: "2021-10-12T14:32:50.913Z",
+        description: "my test group description",
+        group_name: "my test group",
+        last_modified_date: "2021-10-12T14:32:50.913Z",
+        precedence: 19,
+        role_arn: null,
+        user_pool_id: "eu-west-1_Rnma9lp2q",
```
dandwal, line 190: Might be worth considering using a different `user_pool_id` rather than a real one. Having said that I am aware this is exposed elsewhere e.g. The Makefile of the dp-identity-api so just a suggestion but by no means a requirement.

src/app/config/selectors.js
AnnaEmson: Hi Dan, thanks. yes that is correct, I was supposed to change it after talk with James this morning, will update. Thank you

src/app/config/thunks.js
AnnaEmson: One line statement like doesn't need braces. It is valid javascript :-)

dandwal (approved):

PR #681, opened 2022-02-18
The new list Preview teams screen for the new Authentication work

dandwal:
### What

New screen for listing all Preview teams (And roles, there is likely future work to make the roles stand out differently to the preview teams).

This is part of the new Authentication work and talks to Cognito rather than Zebedee for users.

Search functionality:
https://user-images.githubusercontent.com/47502916/154953350-1eb1f896-30b6-49c5-8feb-6920081f3153.mov

Keyboard Navigation:
https://user-images.githubusercontent.com/47502916/154953382-10859565-799a-48d9-8295-1bcf3c4e55fa.mov

Links - (Can preview the links as the cursor hovers look at the bottom left of the screen):
https://user-images.githubusercontent.com/47502916/154953440-58089a14-5675-4986-bffc-a8e058f42ae7.mov

Accessibility score according to google lighthouse:
Screenshot 2022-02-21 at 12 12 13

### How to review

If you would like to run this for yourself then hopefully the instructions below will suffice. Alternatively, I can demo this to you.

To run this we will need to bypass the permissions work (there is still work being done on the backend for this) the instructions below will show you how.
Pull this branch, run the following apps:

- dp-api-router `make debug`
- dp-identity-api  - code changes required before running:
1. Change the [groups endpoint to handler line](https://github.com/ONSdigital/dp-identity-api/blob/8ecb0da7ec1453fbb01b5d9c0b39cdae691e86d6/api/api.go#L112) from
```Go
r.HandleFunc("/v1/groups", auth.Require(GroupsReadPermission, contextAndErrors(api.ListGroupsHandler))).
		Methods(http.MethodGet)
```
To 
```Go
r.HandleFunc("/v1/groups", contextAndErrors(api.ListGroupsHandler)).
		Methods(http.MethodGet)
```
- Florence - code changes required before running:
1. Go into src/index.js change 
```JavaScript
<Route path={`${rootPath}/groups`} component={config.enableNewSignIn ? userIsAuthenticated(userIsAdmin(TeamsList)) : null} />
```
To
```JavaScript
<Route path={`${rootPath}/groups`} component={config.enableNewSignIn ? TeamsList : null} />
```
2. Go to the https://github.com/ONSdigital/florence/blob/11f5fb6fa1d61932f6ec9728bae40abfa6aedd3b/src/app/utilities/api-clients/user.js#L187 and comment out the entire selection statement (if and else part)

Then run Florence with:
`make node-modules; make debug ENCRYPTION_DISABLED=true ENABLE_HOMEPAGE_PUBLISHING=true ENABLE_NEW_SIGN_IN=true ENABLE_DATASET_IMPORT=true`

### Who can review

Anyone except me


src/app/views/teams/teams-view/TeamsList.jsx
```diff
@@ -0,0 +1,103 @@
...
+    const [previewTeams, setPreviewTeams] = useState([]);
+
+    useEffect(() => {
+        loadTeams();
+    }, []);
+
+    useEffect(() => {
+        const listOfPreviewTeams = [];
+        teams?.groups?.forEach(team => {
+            if (team.group_name.startsWith("role-")){
```
AnnaEmson, line 28: it would be good to move this to selectors a that way it will be memorised
so create new selector for example getRoleGroups based on  getGroups  but with that logic inside.

dandwal: This is a really good shout and a good suggestion - was moving it but then after this morning discussion with TL  I have added 'role' type groups back into the list of preview teams so this filter no longer exists or is needed here.

src/app/utilities/api-clients/teams.js
```diff
@@ -2,9 +2,19 @@ import http from "../http";
...
 export default class teams {
     static getAll() {
-        return http.get(`/zebedee/teams`).then(response => {
-            return response.teams;
-        });
+        let config = {};
+        if (window.getEnv != null) {
+            config = window.getEnv();
+        }
+        if (config.enableNewSignIn) {
```
AnnaEmson, line 9: can we start unifying things here? there is api to fetch new groups already created:  https://github.com/ONSdigital/florence/pull/681/files#diff-286ad46cc58e9b11c299bde801fc73666c057e8e68ff57350d8e0691597b1c91R62

src/app/views/teams/teams-view/TeamList.test.jsx
```diff
@@ -0,0 +1,72 @@
+import React from "react";
+import renderer from "react-test-renderer";
+import { render, screen } from "../../../utilities/tests/test-utils";
+import TeamsList from "./TeamsList";
+import "@testing-library/jest-dom/extend-expect";
+
+const teams = {
```
AnnaEmson, line 7: I don't think this is the shape of the object you should get as props. we don't utilise `count`

src/app/views/teams/teams-view/TeamsList.jsx
```diff
@@ -0,0 +1,101 @@
...
+
+    const [isFetchingTeams, setIsFetchingTeams] = useState(true);
+    const [filterTerm, setFilterTerm] = useState("");
+    const [previewTeams, setPreviewTeams] = useState([]);
+
+    useEffect(() => {
+        loadTeams();
+    }, []);
+
+    useEffect(() => {
```
AnnaEmson, line 25: can you please utilise the work around the fetching groups is already implements here: https://github.com/ONSdigital/florence/blob/develop/src/app/config/thunks.js#L146
and used in: https://github.com/ONSdigital/florence/blob/develop/src/app/views/users/groups/AddGroupsToUser.jsx

AnnaEmson: you never set isFetchingTeams to true so the loading state is not visible

src/app/views/teams/teams-view/TeamsList.jsx
```diff
@@ -0,0 +1,101 @@
...
+        const listOfPreviewTeams = [];
+        teams?.groups?.forEach(team => {
```
AnnaEmson, line 27: what is the shape of this teams object?

src/app/views/teams/teams-view/index.jsx
```diff
@@ -0,0 +1,15 @@
+import { connect } from "react-redux";
+import { loadTeamsRequest } from "../../../config/thunks";
+import TeamsList from "./TeamsList";
+
+function mapStateToProps(state) {
+    return {
+        teams: state.state.teams.all,
```
AnnaEmson, line 7: you can use https://github.com/ONSdigital/florence/blob/develop/src/app/config/selectors.js#L29

src/app/views/teams/teams-view/TeamsList.jsx
AnnaEmson: we should not have `team={ groups: []}` but `groups=[]` for now. in the future works on optimising Florence we  will normalise the json object coming form api

AnnaEmson (approved):
nice work Dan! merge it I will unify things so we are good to cut the branch

PR #687, opened 2022-02-25
Update new signin get permissions without email

janderson2:
### What

Update the new signin to get the user's permissions without passing the
`email` query param. The query param method is deprecated with the
migration to JWT auth and it is only possible to check the callers own
permissions based on their access token cookie/header.

### How to review

Ensure that the new sign in is now calling the `GET /permissions` endpoint on zebedee without adding the deprecated `email` query parameter.

### Who can review

Anyone but me.


src/app/utilities/api-clients/user.js
```diff
@@ -109,10 +109,20 @@ export default class user {
...
+     * @deprecated as part of migration to JWT sessions and will be removed soon. Use getPermissions() instead.
+     */
     static getPermissions(email) {
+        console.warn(
+            "WARNING! Deprecated function called. Function 'getPermissions(email)' is deprecated, please use the new 'getPermissions()' instead"
+        );
         return http.get(`/zebedee/permission?email=${email}`, true, true);
     }
 
+    static getPermissions() {
```
dandwal, line 122: I don't think JavaScript is a clever language that can handle function overloading in this manner.

The second declaration will always be used e.g.
https://jsfiddle.net/vw7z9gk8/ (if you open the console in the bottom right window you will see both times the second function will be called).

If this is what you are aiming for then the `getPermissions(email)` is no longer in use we could prob just delete it now

dandwal (approved):

PR #693, opened 2022-03-01
Fixed bug where save broke if topics widget was empty

JoshSinghONS:
### What

Bug fix.

### How to review

LGTM

### Who can review

Anyone.

### DIFF
--- a/src/legacy/js/functions/_t4ArticleEditor.js
+++ b/src/legacy/js/functions/_t4ArticleEditor.js
@@ -167,14 +167,14 @@ function articleEditor(collectionId, data) {
     Florence.globalVars.pagePos = $(".workspace-edit").scrollTop();
 
     // tags
-    if ($("#selectTopic").val() && !$("#selectSubTopic").val()) {
-      sweetAlert("Cannot save this page", "A value is required for 'Subtopic' if a 'Topic' has been selected");
-      return
-    } 
-    else {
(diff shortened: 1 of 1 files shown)


dandwal (approved):

AnnaEmson (approved):
Fabulous! Thank you

src/legacy/js/functions/_t4ArticleEditor.js
```diff
@@ -167,14 +167,14 @@ function articleEditor(collectionId, data) {
...
-      return
-    } 
-    else {
+    if ($("#selectTopic").val() && $("#selectSubTopic").val()) {
       data.description.primaryTopic = $("#selectTopic").val()[0]
       data.description.secondaryTopics = $("#selectSubTopic").val()
     }
+    else if ($("#selectTopic").val() && !$("#selectSubTopic").val()) {
+      sweetAlert("Cannot save this page", "A value is required for 'Subtopic' if a 'Topic' has been selected");
+      return
```
rafahop, line 176: It looks good but how would we feel with something like this instead? I find it would be more readable?
```
if ($("#selectTopic").val()) {
  if ($("#selectSubTopic").val()) {
        data.description.primaryTopic = $("#selectTopic").val()[0]
        data.description.secondaryTopics = $("#selectSubTopic").val()
  } else {
        sweetAlert("Cannot save this page", "A value is required for 'Subtopic' if a 'Topic' has been selected");
        return
  }
}
```

PR #697, opened 2022-03-02
fix: when calendar entry send null not empty string

AnnaEmson:
### What

Before I did refactoring `releaseDateISO` was not set on state at all. 
To avoid component blowing up I needed to add initial value and instead of `null` I created empty string which in consequence is making zebeedee complain and preventing the collection to be created. 
This is putting the previous behaviour back. another question is why do we even use this, I traced history of this file  and it was like that since the app was created. After chatting with backend devs I was advice to just restore previous behaviour and there will be a ticket to investigate that made in future if required.


src/app/views/collections/create/CreateNewCollection.jsx
```diff
@@ -39,7 +39,7 @@ const CreateNewCollection = props => {
     const [newCollection, setNewCollection] = useState(EMPTY_COLLECTION);
     const [isSubmitting, setIsSubmitting] = useState(false);
     const [showScheduleByRelease, setShowScheduleByRelease] = useState(false);
-    const [releaseDateISO, setReleaseDateISO] = useState("");
+    const [releaseDateISO, setReleaseDateISO] = useState(null);
```
rafahop, line 42: I think we could remove this constant completely?

rafahop (commented):
I have tested it locally and it now allows creating a scheduled collection 👍  However I think we could tidy it up and remove the unnecessary constant?

I am also happy that this is reverting the failed attempt to fix it from https://github.com/ONSdigital/florence/pull/695

rafahop (dismissed):

rafahop (approved):

PR #711, opened 2022-03-30
Bug/ unable to add team to collection

AnnaEmson:
### What

Ticket: [Bug/ unable to add team to collection](https://trello.com/c/z1G8RkIm/926-unable-to-add-a-team-to-a-new-collection)

Fix bug created when mapping the groups was removed. 
The change is for old legacy auth so those views will be removed soon.
My solution for this bug is to change the comparison operator to compare values but do not type as old api returns id as a numbers but javascript always sends values in events as a string. 

I also spotted bug on api call so corrected this as well.

Screenshot 2022-03-30 at 14 55 23

before:

Screenshot 2022-03-30 at 15 01 31

### How to review

### Who can review
devs

### DIFF
--- a/src/app/utilities/api-clients/teams.js
+++ b/src/app/utilities/api-clients/teams.js
@@ -46,7 +46,7 @@ export default class teams {
     }
 
     static addMemberToTeam(id, userID) {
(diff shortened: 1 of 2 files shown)


mr-nick17 (approved):
LGTM

PR #722, opened 2022-04-04
Added language toggle to nav bar

JoshSinghONS:
### What

Added language toggle to nav bar - https://trello.com/c/mKNxFppq/712-404-error-when-language-toggle-is-clicked-in-preview-within-florence

Rather than try and get the language toggle within the preview to function which was proving difficult due to CORS and it existing within an iframe it was decided to implement a button similar to what a publisher/admin account has access to in the create/edit UI for viewer accounts or those simply previewing a collection.

### How to review

Look at the code.

### Who can review

Those with React experience.

### DIFF
--- a/src/app/components/navbar/NavBar.jsx
+++ b/src/app/components/navbar/NavBar.jsx
@@ -1,12 +1,15 @@
-import React from "react";
+import React, { useState } from "react";
 import { Link } from "react-router";
 import PropTypes from "prop-types";
 import url from "../../utilities/url";
 import auth from "../../utilities/auth";
 import user from "../../utilities/api-clients/user";
+import cookies from "../../utilities/cookies";
 import PreviewNav from "../preview-nav";
 
 const NavBar = props => {
+    const [lang, setLang] = useState(cookies.get("lang"));
+
     const renderWorkingOnItem = () => {
         const workingOn = props.workingOn || {};
         const showWorkingOn = workingOn.id;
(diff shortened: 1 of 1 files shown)


src/app/components/navbar/NavBar.jsx
```diff
@@ -36,6 +39,19 @@ const NavBar = props => {
...
+        if (lang == "en") {
+            cookies.remove("lang");
+            cookies.add("lang", "cy", null);
+            setLang("cy");
+        } else {
+            cookies.remove("lang");
+            cookies.add("lang", "en", null);
+            setLang("en");
+        }
+        window.location.reload(true);
```
AnnaEmson, line 52: I don't think you need this.

src/app/components/navbar/NavBar.jsx
```diff
@@ -1,12 +1,15 @@
...
 import { Link } from "react-router";
 import PropTypes from "prop-types";
 import url from "../../utilities/url";
 import auth from "../../utilities/auth";
 import user from "../../utilities/api-clients/user";
+import cookies from "../../utilities/cookies";
 import PreviewNav from "../preview-nav";
 
 const NavBar = props => {
+    const [lang, setLang] = useState(cookies.get("lang"));
```
AnnaEmson, line 11: I would be better to read the cookie on component load. 

```
    useEffect(() => {
        setLang(cookies.get("lang"))
    }, []);

```

AnnaEmson (commented):
Would it be possible to upload a screenshot or video how it is working as ticket doesn't have AC for this solution? 
But from the  code I can see you added new menu item and it is switching the language of the preview?  would it be useful to update styling to show that it is a button and / or give some idea of the action here like text ? `Switch Preview to English`

src/app/components/navbar/NavBar.jsx
AnnaEmson: Also I thought this is the change of language for whole florence, maybe you can call it in here 
`const [previewLanguage, setPreviewLanguage] = useState('');`

src/app/components/navbar/NavBar.jsx
```diff
@@ -36,6 +39,19 @@ const NavBar = props => {
         return path.indexOf(`/collections`) >= 0;
     };
 
+    const changeLang = () => {
```
AnnaEmson, line 42: `changePreviewLanguage` ?

AnnaEmson (approved):

PR #730, opened 2022-04-05
ability to remove Interactive - ticket number: #98

geoffmrb:
commit 8156b0f795c22e64f4aa3b09747d74e0b84d029b
Author: geoffmrb <g_ofori@hotmail.co.uk>
Date:   Tue Apr 5 08:27:13 2022 -0500

    Delete an interactive page

### What

ability to remove an interactive after it has been added

https://user-images.githubusercontent.com/58171257/162009931-a25e669b-7095-4a8b-9f15-45e3e4bf82de.mp4

### How to review

access interactives, add one then please remove it from the index screen

### Who can review

https://trello.com/c/BeY9dIPk/98-add-delete-control-to-the-interactives-screens


AnnaEmson (changes requested):
I stop reviewing as  there are already lots of comments and difficult go thru this. Please correct and this can be reviewed again.
Just few general points:

Florence is now updated to react 17.02 so you can use functional components and hooks. Those will be clearer and easier to maintain. 

Respect current folders structure within Florence. 
Don't call views controllers. You can see those in Florence but we are refactoring this.
Don't keep few components in one file, please extract them to separate folders structure. Make use of components already existing in Florence. This the approach we are taking in components so this can be easily maintained .  

The code should be divided into separate components so they can be reused for better maintainability and consistency.
There is no ticket in the description so very difficult to judge the scope nor images or video to see this inaction.

We should avoid using ids in the css or !. Your changes are targeting other parts of Florence which I don't think should be happening. 
You are not using of the styling or elements already functioning gin Florence.

You do not need to use id when targeting elements in react testing library 
you should test events and changes in UI not implantation I can't see this in your tests.
more as per principles :

`In the spirit of [the guiding principles](https://testing-library.com/docs/guiding-principles), it is recommended to use this only after the other queries don't work for your use case. Using data-testid attributes do not resemble how your software is used and should be avoided if possible. That said, they are way better than querying based on DOM structure or styling css class names. Learn more about data-testids from the blog post ["Making your UI tests resilient to change"](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change)`

PR #733, opened 2022-04-11
Added language toggle for Florence preview

JoshSinghONS:
### What

Added a language toggle in the nav bar so that the preview page in Florence can display in welsh.

### How to review

Sanity check the code / preview a collection locally and flip the toggle.

### Who can review

Anyone


JoshSinghONS:
Hi @AnnaEmson,

Context: Language change button on collection preview breaking select box nav.

Been trying to fix this and I thought it might just be easier, quicker and hopefully a learning opportunity for me if you could look over the code and maybe steer my wandering ship?

So since windows.reload() breaks the select box because it erases the current state I thought the best option forward was for NavBar component when the button is clicked to tell the iframe to do the update however it seemed a bit tedious to work my way through the tree via listeners and emits so I looked at doing it via state.

I've given it my best go between looking at what is currently there and the redux guide but I've fallen short when it comes to getting the iFrame to watch the state, I was initially going to use store.subscribe but then I thought mapstatetoprops() might the answer?

TLDR: Have I set up the languagesReducer correctly? How should I get the Iframe component to listen to and react to it?

src/app/components/navbar/NavBar.jsx
```diff
@@ -8,34 +10,32 @@ import cookies from "../../utilities/cookies";
 import PreviewNav from "../preview-nav";
 
 const LANG = {
-    "en": "English",
-    "cy" : "Welsh"
-}
+    en: "English",
+    cy: "Welsh",
+};
```
JoshSinghONS, line 15: I can probably delete this now right if I have languageConstants?

src/app/components/navbar/NavBar.jsx
```diff
@@ -8,34 +10,32 @@ import cookies from "../../utilities/cookies";
...
 
 const NavBar = props => {
     const [previewLanguage, setPreviewLanguage] = useState(null);
```
JoshSinghONS, line 18: Can also delete this if there is a global state?

src/app/components/navbar/NavBar.jsx
```diff
@@ -69,10 +69,11 @@ const NavBar = props => {
     const changeLang = () => {
         if (previewLanguage == "en") {
             setPreviewLanguage("cy");
+            store.dispatch(setNotEnglish());
```
JoshSinghONS, line 72: So ideally in my head that is going to the store and using the SET_NOT_ENGLISH action found in the languageReducer, is that correct?

AnnaEmson:
@JoshSinghONS  will look at this and will update you with my opinion

JoshSinghONS:
> @JoshSinghONS very good thinking, and that is the best way to do it. will look at this and will update you with my opinion

Thank you, the state seems to be recorded and the iFrame seems to be able to see it and it records it in its props but when I try and log it it keeps saying undefined and doesn't seem to want to re-render even though it should :/ hence all the console logs lol

AnnaEmson:
Hi @JoshSinghONS I updated you the commit. well done so far. Few improvements. Now you have language in iframe. Not sure what you are planning to do with the cookie though. I comment is for now. Are you pllaning to set language on cookie to and sync it with state?

src/app/components/navbar/NavBar.jsx
AnnaEmson: I would probably leave it as it is a dictionary for yours abbreviations. Maybe we don't need abbreviations?

src/app/components/navbar/index.jsx
```diff
@@ -1,13 +1,20 @@
...
+import { setPreviewLanguage } from "../../config/actions";
 import NavBar from "./NavBar";
 
 const mapStateToProps = state => ({
     user: state.user,
     rootPath: state.state.rootPath,
     workingOn: getActive(state.state),
     config: state.state.config,
-    isNewSignIn: getEnableNewSignIn(state.state),
+    previewLanguage: getPreviewLanguage(state.state),
```
JoshSinghONS, line 11: Bit confused on the state syntax here. 

What exactly is state versus state.state?

Why is user state.user and not state.state.user?

src/app/components/navbar/NavBar.jsx
```diff
@@ -13,27 +13,8 @@ const LANG = {
...
-            changeLanguageCookie(previewLanguage);
-        }
-    }, [previewLanguage, changeLanguageCookie]);
-
-    const changeLanguageCookie = previewLanguage => {
-        cookies.remove("lang");
-        cookies.add("lang", previewLanguage, null);
-    };
+    console.log("previewLanguage", props.previewLanguage);
+    useEffect(() => props.setPreviewLanguage("en"), []);
```
JoshSinghONS, line 17: initialState.js has previewLanguage set to "en" 

Does this mean I can remove the useEffect?

JoshSinghONS:
> Hi @JoshSinghONS I updated you the commit. well done so far. Few improvements. Now you have language in iframe. It is also updated if you click on the button. you can see in console log I left for you. Not sure what you are planning to do with the cookie though. I commented it for now. Are you planning to set language on cookie too and sync it with state or this will be removed? I removed your changes in Iframe component as old structure is the PreviewController holds connection to redux and Iframe is treated here as a view child component and I think for now we should keep it that way. This can be refactored later, you probably don't want to do it at this stage

Hi @AnnaEmson great stuff! 

Yeah so my idea was to set the cookie and the state prop with the latter telling the iFrame if it needs to re-render and when it does the cookie change will cause it to be in welsh.

src/app/components/navbar/index.jsx
AnnaEmson: ok, so it is because someone designed this this way. it call comes form the store.js and this line:
https://github.com/ONSdigital/florence/blob/develop/src/app/config/store.js#L19
you can see users are in separate reducer and this is why coming under their name. 
Ideally I would like to change that just seems very tricky before refactor

src/app/components/navbar/NavBar.jsx
AnnaEmson: something need to transport it to here and this is what the useEffect does here. we can use useSelector if you don't like this way

src/app/components/iframe/Iframe.jsx
```diff
@@ -37,7 +37,7 @@ export default class Iframe extends Component {
     }
 
     render() {
-        return <iframe id="iframe" className="preview__iframe" src={this.props.path}></iframe>;
+        return <iframe id="iframe" key={this.props.previewLanguage} className="preview__iframe" src={this.props.path}></iframe>;
```
JoshSinghONS, line 40: I don't like having to use key, feels hacky but not sure how else I can get it to behave the way I want.

AnnaEmson:
@JoshSinghONS this PR is `out-of-date with the base branch and please change the title as it implies this is still WIP and also fill the PR template :-)

AnnaEmson (approved):

PR #734, opened 2022-04-11
Set interactives flag

geoffmrb:
### What

### How to review

### Who can review

### DIFF
--- a/src/app/components/navbar/NavBar.test.js
+++ b/src/app/components/navbar/NavBar.test.js
@@ -11,6 +11,7 @@ const authenticatedViewer = createMockUser("user@test.com", true, true, "VIEWER"
 const defaultProps = {
     config: {
         enableDatasetImport: false,
+        enableNewInteractives: false,
     },
     user: notLoggedUser,
     rootPath: "/florence",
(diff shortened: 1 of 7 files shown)


src/app/components/navbar/NavBar.test.js
```diff
@@ -90,18 +91,18 @@ describe("NavBar", () => {
             });
         });
 
-        describe("when enabled dataset import", () => {
-            it("should display Datasets", () => {
+        describe("when enabled upload interactives", () => {
```
AnnaEmson, line 94: Can you please revert back your changes as we would like to keep this test and create new test case for your feature.  enabled dataset import feature flag is still active

src/app/components/navbar/NavBar.test.js
```diff
@@ -46,7 +47,7 @@ describe("NavBar", () => {
         it("should render navigation with links", () => {
             const component = shallow(<NavBar {...defaultProps} user={authenticatedUser} />);
             const nav = component.find(Link);
-
+
```
AnnaEmson, line 50: trailing whitespaces, worth before opening PR run lint so your code is formatted and those errors removed

src/app/components/navbar/NavBar.test.js
geoffmrb: there was a duplicate with this, so only replaced the duplicate

PR #740, opened 2022-04-25
Replace gorilla/pat router with gorilla/mux

jessjenkins:
### What

A security vulnerability in the `gorilla/pat` router allows a maliciously crafted url to act as an open redirect. Although florence is not available externally, an attacker could theoretically craft a malicious redirect in an emailto an internal staff member. `gorilla/mux` does not suffer from the same vulnerability as it correctly escapes the url supplied to the Location header when performing a 301 redirect.

### How to review

Ensure the service no longer uses `gorilla/pat` for routing and that an appropriate test exists and passes.

### Who can review

Anyone but me


justinpjose (approved):
Looks good and tests passes

PR #741, opened 2022-04-26
New Auth - corrections

AnnaEmson:
### What

This PR is the update for new auth functionality: 
1. Bug - when creating police pass groups id not collection
2. show list of new groups in Collection Edit when NewSignIn flag is on
3. When creating policy use put so policy id is the same as collection id
4. create api calls
5. include action when editing user as was not included 
6. #945: Users tab should only be visible to admins

### How to review

### Who can review

### DIFF
--- a/src/app/config/selectors.js
+++ b/src/app/config/selectors.js
@@ -7,7 +7,7 @@ export const getCollections = state => state.collections.all;
 export const getSearch = state => state.search;
 
 export const getMappedCollections = createSelector(getCollections, collections => {
-    if (!collections) return [];
+    if (!collections || collections.length === 0 || typeof collections === "string") return [];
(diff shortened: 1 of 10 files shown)


src/app/config/selectors.js
```diff
@@ -7,7 +7,7 @@ export const getCollections = state => state.collections.all;
 export const getSearch = state => state.search;
 
 export const getMappedCollections = createSelector(getCollections, collections => {
-    if (!collections) return [];
+    if (!collections || collections.length === 0 || typeof collections === "string") return [];
```
AnnaEmson, line 10: this is temporary as collection api returns string with error message you are not authenticated. If this is removed the app crashes

mr-nick17 (approved):

PR #754, opened 2022-05-06
interactive preview: Rename update to interactive

geoffmrb:
### What

amended endpoint of api, to match preview 

### How to review

look at the preview page on an edit interactive screen

### Who can review

anyone


mryan321 (approved):
lgtm - matches what api expects

PR #762, opened 2022-05-11
Feature: Update packages: 1

James-Nitron:
### What

Updated outdated packages, and ran an npm audit fix. Confirmed that the app seems to work on node 17. 
Installed URL, needed for react-router-redux.

### How to review

Fetch, and run the app.

### Who can review

Anyone.


src/package.json
```diff
@@ -112,13 +113,13 @@
...
-    "redux": "^3.7.2",
+    "redux": "^4.2.0",
     "redux-auth-wrapper": "^2.0.3",
-    "redux-devtools-extension": "^2.13.7",
-    "redux-thunk": "^2.3.0",
+    "redux-thunk": "^2.4.1",
     "reselect": "^4.1.4",
     "sass": "^1.34.1",
     "ssri": "^6.0.2",
+    "url": "^0.11.0",
```
gilesburdett, line 122: Why does the `url` package need to be introduced as part of the upgrade? It seems to very popular but at the same time hasn't been updated for seven years, and Node has surely moved on in that time with its own implementation.

designorant (changes requested):

PR #769, opened 2022-05-16
Added check boxes to required content types

JoshSinghONS:
### What

Added census check box to various content types.
Unify checkbox handling (see https://github.com/ONSdigital/florence/pull/773)

### How to review

Sanity.

### Who can review

Anyone.

### DIFF
--- a/src/legacy/js/functions/_t16ReleaseEditor.js
+++ b/src/legacy/js/functions/_t16ReleaseEditor.js
@@ -115,6 +115,10 @@ function releaseEditor(collectionId, data) {
         
     });
 
+    $("#survey-checkbox").click(function () {
+        data.description.survey = $("#survey-checkbox").prop('checked') ? 'census' : null;
+    });
+
     $("#cancelled input[type='checkbox']").prop('checked', checkBoxStatus($('#cancelled').attr('id'))).click(function () {
         data.description.cancelled = $("#cancelled input[type='checkbox']").prop('checked') ? true : false;
         if (data.description.cancelled) {
--- a/src/legacy/js/functions/_t16ReleaseEditor.js
+++ b/src/legacy/js/functions/_t16ReleaseEditor.js
@@ -86,41 +86,26 @@ function releaseEditor(collectionId, data) {
         
     });
 
-    /* The checked attribute is a boolean attribute, which means the corresponding property is true if the attribute is present at all—even if, for example, the attribute has no value or is set to empty string value or even "false" */
-    var checkBoxStatus = function (id) {
-        if (id === 'natStat') {
-            if (data.description.nationalStatistic === "false" || data.description.nationalStatistic === false) {
-                return false;
-            } else {
-                return true;
-            }
-        } else if (id === 'cancelled') {
-            if (data.description.cancelled === "false" || data.description.cancelled === false) {
-                return false;
-            } else {
-                return true;
-            }
-        } else if (id === 'finalised') {
-            if (data.description.finalised === "false" || data.description.finalised === false) {
-                return false;
-            } else {
-                return true;
-            }
+    /* The checked attribute is a boolean attribute and the corresponding property 
+    will be true if the attribute is present and has a value other than false */
+    var checkBoxStatus = function (value) {
+        if (value === "" || value === "false" || value === false) {
+            return false;
         }
+        return true;
     };
 
     // Gets status of checkbox and sets JSON to match
-    $("#natStat input[type='checkbox']").prop('checked', checkBoxStatus($('#natStat').attr('id'))).click(function () {
-        data.description.nationalStatistic = $("#natStat input[type='checkbox']").prop('checked') ? true : false;
-        
+    $("#natStat-checkbox").prop('checked', checkBoxStatus(data.description.nationalStatistic)).click(function () {
+        data.description.nationalStatistic = $("#natStat-checkbox").prop('checked');
     });
 
-    $("#census").click(function () {
-        data.description.survey = $("#census").prop('checked') ? 'census' : null;
+    $("#census-checkbox").prop('checked', data.description.survey ? true : false).click(function () {
+        data.description.survey = $("#census-checkbox").prop('checked') ? 'census' : null;
     });
 
-    $("#cancelled input[type='checkbox']").prop('checked', checkBoxStatus($('#cancelled').attr('id'))).click(function () {
-        data.description.cancelled = $("#cancelled input[type='checkbox']").prop('checked') ? true : false;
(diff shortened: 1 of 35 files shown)


src/legacy/js/functions/_t4BulletinEditor.js
```diff
@@ -101,6 +101,10 @@ function bulletinEditor(collectionId, data) {
         data.description.nationalStatistic = $("#metadata-list input[type='checkbox']").prop('checked') ? true : false;
     });
 
+    $("#survey-checkbox").prop('checked', data.description.survey ? true : false).click(function () {
```
rafahop, line 104: No need for the ternary operator here?
```suggestion
    $("#survey-checkbox").prop('checked', data.description.survey).click(function () {
```

Also should this prop be included in articles and releases too? It looks like we do them in the handlebars template rather than js for them. We should try and be consistent. Which one would be a better practice?

src/legacy/templates/workEditT16.handlebars
```diff
@@ -82,6 +82,10 @@
                         <label for="cancelled">Cancelled </label>
                         <input type="checkbox" name="cancelled" value="false" class="checkbox"/>
                     </div>
+                    <div id="survey">
+                        <label for="survey-checkbox">Census </label>
+                        <input id="survey-checkbox" type="checkbox" name="survey" {{#if this.description.survey}}checked{{/if}}/>
```
rafahop, line 87: A bit picky here but the naming is not very accurate. This is the census checkbox. But we also talked about this potentially evolving to a survey dropdown and made the backend change accordingly (hence the name of the field being survey)

My view is that this should be just about census for now to avoid any confusion. If we were to add surveys in the future, this would need to change anyway as there won't be a survey checkbox so better name it accurately now?

src/legacy/js/functions/_t4BulletinEditor.js
JoshSinghONS: If we don't have the ternary won't it just try and inject the value of .survey into the checked property when it only accepts true or false? 

So if something has the census brand attached to it won't it try and inject the string "census" and if it does not have the census brand won't it try and inject null or possible undefined?

In terms of the prop being included for articles and releases I actually tried to remain consistent by having the bulletin and release pages handle the checking in the handlebars instead but they refused, why I have no idea as I could see description.survey in the template data when it passed through.

I could change articles and releases to instead use the prop but my thinking was that each file remains consistent with itself which I think is less confusing. 

For example if I change articles and releases to have the changes bulletins and datasets have they will then have the national statistic checkbox property decided in the handlebars but the census branding checkbox decided by the prop which I think in it's own way is more confusing.

src/legacy/templates/workEditT16.handlebars
JoshSinghONS: Fair

src/legacy/js/functions/_t4BulletinEditor.js
rafahop: You are absolutely right! Not sure why I thought `data.description.survey` contained a boolean... Yes, the ternary operator is needed, sorry for the confusion!

Re the checked props, it all seems odd! I have been playing with it and I can't get the articles or releases to work as is. Not sure what is different to the national statistics checkbox that seem to work. Maybe we should go to the bottom of this? 
I feel like we should try and make it consistent across files (work with all checkbox in the same way) even if that means a file might start having inconsistencies with itself? But open to discuss it

rafahop: Just noticed this changes would make the release file not be consistent with itself either! 🤷

rafahop: Made the handling of checkboxes consistent across files (https://github.com/ONSdigital/florence/pull/773)

src/legacy/js/functions/_t16ReleaseEditor.js
```diff
@@ -87,36 +87,24 @@ function releaseEditor(collectionId, data) {
...
-            } else {
-                return true;
-            }
-        } else if (id === 'finalised') {
-            if (data.description.finalised === "false" || data.description.finalised === false) {
-                return false;
-            } else {
-                return true;
-            }
+    var checkBoxStatus = function (value) {
```
cookel2, line 90: The comment above this line does not seem to agree with what this function is doing. I might be misunderstanding it but it seems to suggest that if the value is "" or "false" then it should return true?

src/legacy/js/functions/_t4ArticleDownloadEditor.js
```diff
@@ -78,15 +78,15 @@ function ArticleDownloadEditor(collectionId, data) {
 
   /* The checked attribute is a boolean attribute, which means the corresponding property is true if the attribute
    is present at all—even if, for example, the attribute has no value or is set to empty string value or even "false" */
-  var checkBoxStatus = function () {
-    if (data.description.nationalStatistic === "false" || data.description.nationalStatistic === false) {
+  var checkBoxStatus = function (value) {
```
cookel2, line 81: Same question about the comment on line 79-80 as above.

cookel2 (commented):
A couple of small comments about comments.

src/legacy/templates/workEditT16.handlebars
```diff
@@ -82,6 +82,10 @@
                         <label for="cancelled">Cancelled </label>
                         <input type="checkbox" name="cancelled" value="false" class="checkbox"/>
                     </div>
+                    <div id="census">
+                        <label for="census-checkbox">Census </label>
+                        <input id="census-checkbox" type="checkbox" name="census" />
```
cookel2, line 87: Should line 87 also have class="checkbox" similar  to line 83?

src/legacy/js/functions/_t16ReleaseEditor.js
rafahop: Yes, you are right! The comment does not seem to make much sense. It is present in several files and I have copied it over without thinking about it too much. I will amend them!

src/legacy/templates/workEditT4Article.handlebars
```diff
@@ -77,7 +77,11 @@
                         </div>
                         <div id="natStat">
                             <label for="natStat-checkbox">National statistic</label>
-                            <input id="natStat-checkbox" type="checkbox" name="natStat" {{#if this.description.nationalStatistic}}checked{{/if}}/>
+                            <input id="natStat-checkbox" type="checkbox" name="natStat" />
```
cookel2, line 80: Does it need class="checkbox" here too?
```suggestion
                            <input id="natStat-checkbox" type="checkbox" name="natStat" class="checkbox"/>
```

cookel2 (commented):
Just one more small thing.

cookel2 (approved):

PR #774, opened 2022-05-18
Fix for conflicting cookie

JoshSinghONS:
### What

Set domain for cookie to fix conflict.

### How to review

Sanity

### Who can review

Anyone

### DIFF
--- a/src/legacy/js/functions/_createWorkspace.js
+++ b/src/legacy/js/functions/_createWorkspace.js
@@ -28,9 +28,9 @@ function createWorkspace(path, collectionId, menu, collectionData, stopEventList
 
         Florence.globalVars.pagePath = safePath;
         if (Florence.globalVars.welsh !== true) {
-            document.cookie = "lang=" + "en;path=/";
+            document.cookie = "lang=" + "en;path=/" + ";domain=" + location.hostname;
         } else {
-            document.cookie = "lang=" + "cy;path=/";
+            document.cookie = "lang=" + "cy;path=/" + ";domain=" + location.hostname;
         }
         Florence.refreshAdminMenu();
 


src/legacy/js/functions/_createWorkspace.js
```diff
@@ -28,9 +28,9 @@ function createWorkspace(path, collectionId, menu, collectionData, stopEventList
 
         Florence.globalVars.pagePath = safePath;
         if (Florence.globalVars.welsh !== true) {
-            document.cookie = "lang=" + "en;path=/";
+            document.cookie = "lang=" + "en;path=/" + ";domain=" + location.hostname;
         } else {
-            document.cookie = "lang=" + "cy;path=/";
+            document.cookie = "lang=" + "cy;path=/" + ";domain=" + location.hostname;
```
rafahop, line 33: Could we avoid duplication here please? If the only difference is the language value we could use a local variable.

Also there is no need to concatenate 2 strings?
```suggestion
            document.cookie = "lang=" + "cy;path=/;domain=" + location.hostname;
```

rafahop (changes requested):

src/legacy/js/functions/_createWorkspace.js
```diff
@@ -27,11 +27,14 @@ function createWorkspace(path, collectionId, menu, collectionData, stopEventList
...
         Florence.globalVars.pagePath = safePath;
+        var languageCookie = "lang="
         if (Florence.globalVars.welsh !== true) {
-            document.cookie = "lang=" + "en;path=/";
+            languageCookie += "en"
         } else {
-            document.cookie = "lang=" + "cy;path=/";
+            languageCookie += "cy"
         }
+        languageCookie += ";path=/;domain=" + location.hostname;
```
rafahop, line 36: This is has nothing to do with a language cookie if we append the path and domain. Maybe rename the variable or just do this instead?

```
document.cookie = languageCookie + ";path=/;domain=" + location.hostname;
```

rafahop (approved):

PR #781, opened 2022-05-20
Remove package-lock.json

rafahop:
### What

Remove `package-lock.json` as this shouldn't live in the root folder
Remove entry in `.gitignore` which incorrectly tried to ignore this file in the `src` file when it should be checked in

### How to review

Check changes make sense

### Who can review
Anyone


Marianasilvago (approved):
LGTM

PR #792, opened 2022-05-31
Interactives: link to collection updates and some publishing workflow work

geoffmrb:
### What

hook up to api endpoints to update collection and interactive link

image
image

### How to review

collections > create/edit content > interactives > link

### Who can review

anyone

### DIFF
--- a/src/app/utilities/api-clients/collections.js
+++ b/src/app/utilities/api-clients/collections.js
@@ -159,4 +159,29 @@ export default class collections {
             return response;
         });
     }
+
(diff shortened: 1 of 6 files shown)


joegasewicz (approved):

rafahop:
@geoffmrb can we please ensure that we run florence locally to make sure the postinstall is executed meaning we check in changes with the right formatting please? `make dev` would do. This will avoid having to keep raising PRs like https://github.com/ONSdigital/florence/pull/797

PR #797, opened 2022-06-01
Updaate files touched by prettier in postinstall

rafahop:
### What

Commit files touched by `prettier` in the postinstall

When running `make node-modules-react` on a fresh copy of the repository, `prettier` touches several files as part of the postinstall task. This leaves a dirty copy of the repository that cannot be updated without clearing out the changes.

### How to review

- Run `make node-modules-react`
- Verify that no files tracked by git are changed

### Who can review

Anyone


rafahop:
No longer required. It has been fixed in subsequent commits

PR #801, opened 2022-06-02
Interactives - Remove interactive from collection

geoffmrb:
### What

gives ability to user to remove interactive from collection

### How to review

try to remove interactive from collection

### Who can review

anyone


mryan321 (approved):
im not even sure this is valid now but ill approve it because its just within our interactives stuff - the delete ticket was talking about deleting from here - from a collection -  i dont think this covers it:

Screenshot 2022-06-02 at 18 25 05

mryan321 (approved):

PR #821, opened 2022-06-08
Fix the refactored watcher

designorant:
### What

Fixed `watch-refactored` so `make watch-src` works as intended and rebuilds the files with changes.

I really wanted to add HMR or at least LiveReload in there but it doesn't seem that we actually use the webpack development server in serve mode, likely due to the convoluted structure and port mappings. There seems to be leftovers from webpack v4 config in there but that's a whole different story: there's a lot in there that could do with a solid cleanup, but this must suffice for now.

### How to review

0. Fire up `florence`
1.  Run `make watch-src`
2. Edit a legacy / old workspace file, refresh, see rebuilt changes
3. Edit a refactored / new workspace file, refresh, see rebuilt changes

### Who can review

Anyone


designorant:
Rebased on latest `develop`.

designorant:
Rebased and added two prettify files caught by the watcher.

gilesburdett (approved):

PR #835, opened 2022-06-16
put banner for sandbox (was just in develop or sandpit)

gedge:
### What

in florence
we used to have a banner for `You're on develop env`
and this ensures we now get `You're on sandbox env`


redhug1 (approved):
LGTM

PR #836, opened 2022-06-16
#986: Check for existing auth state on login

joegasewicz:
Screenshot 2022-06-20 at 17 39 17

### What

What
------

* Update initial florence load logic as outlined below

On original Florence load we will need to:

1. Check local state to get `ons_auth_state` and the expiry times within
2. If access token has expired (based on time in auth state object then refresh it
3. Start timers 1 and 2 as listed below

N.B. make sure that the above logic works with:

* both current and JWT based sessions
* legacy and new world (React) florence

The session refresh logic has 3 timers:

1. Access token expiry count down timer
2. Refresh token count down timer
3. Time since last access (no longer needed)

EXTRA:

1. Make sure current florence refresh interval is set to 12 hours not 24.

### How to review

Read through the code, the main thing to think about is that we don't cause any issues with flashing up the warning constantly which could cause issues for other teams.

You can pull down the code and step through it if you have the time but I don't expect you to do that due to the complexity of this feature
### Who can review

Anyone familiar with florence
myself


src/app/config/user/userHooks.js
```diff
@@ -1,21 +1,55 @@
...
             });
         } else {
-            setUserState(userToken);
+            setShouldUpdateAccessToken(false);
+            setUserState(authState);
         }
     }, []);
     return userState;
 };
+
```
joegasewicz, line 28: This hook makes it very easy to remove the timers from the Layout component should there be any issues

mr-nick17 (approved):
LGTM - demoed and checked login still works locally

mr-nick17 (approved):

PR #841, opened 2022-06-20
Format files

rafahop:
### What

[prettier](https://www.npmjs.com/package/prettier) is run as part of the build making these files show as changed locally even when we haven't touched them. Checking in the right version of them so that they don't annoy developers

### How to review

Check changes are only format changes. 
Run `make dev` and see there are no local changes. 

### Who can review

Anyone


cookel2 (approved):
looks good to me

PR #846, opened 2022-06-22
#983: Remove access token expiry pop-up

joegasewicz:
### What

- Remove the "Your session will expire" pop up as isn't adding value ✅ 
  - Pop up isn't adding any user value since their session isn't actually about to expire it is just the access token that is ✅ 
- What needs changing:
  * Remove last interaction/active timer (3 below) in session refresh login in Florence ✅ 
  * Update the "auth state" with the new access token expiry time ✅ 
  * Replace current session refresh logic (incl. the popup it produces) with logic that just before access token expiry (maybe 1 min before) based on access token expiry timer (1 below), refresh the access token (i.e. `PUT /api/v1/tokens/self`) and then reset timer ✅ 

### How to review
Check the code does meets the above expectations

### Who can review
Forence devs

myself


mr-nick17 (approved):
LGTM

PR #848, opened 2022-06-24
florence legacy auth updates

joegasewicz:
### What

**_ALL UPDATES IN THIS PR ARE FOR FLORENCE LEGACY_**  

* Update initial florence load logic as outlined below

On original Florence load we will need to:

1. Check local state to get `ons_auth_state` and the expiry times within ✅ 
2. If access token has expired (based on time in auth state object then refresh it ✅ 
3. Start timers 1 and 2 as listed below ✅ 

N.B. make sure that the above logic works with: ✅ 

The session refresh logic has 3 timers: ✅ 

1. Access token expiry count down timer ✅ 
3. Time since last access (no longer needed) ✅ 

EXTRA:

1. Make sure current florence refresh interval is set to 12 hours not 24. ✅ 
* Ensure that any 401 response from a resource API then tries to refresh the access token by calling `PUT /api/v1/tokens` on dp-identity-api ✅ 
 
When 401 received from resource API: ✅ 

* On any 401 response from resource APIs, clear the "auth state" from local storage and redirect to login screen with `redirect` query param set ✅ 

* If new sign in feature flag enabled then attempt to refresh token: `PUT /api/v1/tokens/self` on dp-identity-api: ✅ 
  * If successful then: ✅
    1. Update the access token expiry time in auth state ✅
    2. Reset access token expiry timer ✅
    3. Retry original request (up to retry limit) ⛔  (not required)
  * If receive a 401 from token refresh endpoint then:  ✅
    1. Clear the "auth state" from the local storage ✅
    2. Redirect the user to the login screen with the `redirect` query param set ✅
* Else: (same as 401 from token refresh in above block) ✅
    1. Clear the "auth state" from the local storage ✅
    2. Redirect the user to the login screen with the `redirect` query param  ✅

- Remove the "Your session will expire" pop up as isn't adding value ⛔ (required in legacy please ask for reason)
  - Pop up isn't adding any user value since their session isn't actually about to expire it is just the access token that is ⛔  (see above)
- What needs changing:
  * Remove last interaction/active timer (3 below) in session refresh login in Florence ⛔ 
  * Update the "auth state" with the new access token expiry time ✅ 
  * Replace current session refresh logic (incl. the popup it produces) with logic that just before access token expiry (maybe 1 min before) based on access token expiry timer (1 below), refresh the access token (i.e. `PUT /api/v1/tokens/self`) and then reset timer ⛔ 

The session refresh logic has 3 parts:

1. Access token expiry count down timer
2. Refresh token count down timer
3. Time since last access (no longer needed)

N.B. make sure that the above logic works with:

* both current and JWT based sessions ✅ 
* legacy and new world (React) florence ✅ 

### How to review

Either read through the code or request a local demo from myself.

As above

### Who can review

Florence Frontend devs
Myself

### DIFF
--- a/src/legacy/js/functions/_sessionManagement.js
+++ b/src/legacy/js/functions/_sessionManagement.js
@@ -1,3 +1,24 @@
+// Session Management
+// Warning: for testing only
+function __TEST_TIMERS(name, match, i) {
+    const minutes = 60 * 1000;
+    if (name === match) {
+        const now = new Date();
+        // interval will be less than a minute
+        const interval = now.setTime(now.getTime() + -57 * minutes);
+        console.debug("[TEST] Updating " + name + " interval of " + new Date(interval));
+        return new Date(interval);
+    }
+    return i;
+}
+
+function timeOffsets() {
+    return {
+        passiveRenewal: 300000,
+        invasiveRenewal: 120000,
+    };
+}
+
 // There are two tokens, Session and Refresh that manage the users access. Refresh is a long life token that can be
 // used to get a new session token. The session token is what gives the user access to the system and has a very
 // short life. If the user is active then we extend their session. The refresh token has a long life but also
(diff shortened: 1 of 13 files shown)


src/legacy/js/functions/_sessionManagement.js
```diff
@@ -115,14 +140,111 @@ function refreshSession() {
...
+            console.error(err);
+        });
+        return res;
+}
+
+/**
+ *  @param sessionExpiryTime must be in js date format (not UTC from server)
+ *  @returns true if session has expired
+ *  */
+function isSessionExpired(sessionExpiryTime, ) {
```
joegasewicz, line 200: This might need to be updated in current florence

geoffmrb (dismissed):

geoffmrb (dismissed):

geoffmrb (approved):

PR #889, opened 2022-08-04
Fix/enable parity

JoshSinghONS:
### What

Added commits from master which were not in develop and fixed merge conflicts.

### How to review

LGTM.

### Who can review

Anyone.

### DIFF
--- a/src/audit-allowlist.json
+++ b/src/audit-allowlist.json
@@ -27,6 +27,18 @@
         {
             "id": "CVE-2022-2596",
             "reason": "Several libraries appear locked to node-fetch@2 & cannot bump to node-fetch@3.2.10"
+        },
+        {
+            "id" : "CVE-2022-24785",
+            "reason" : "Moment.js"
+        },
+        {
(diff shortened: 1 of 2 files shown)


rafahop:
Aren't these different commits to the ones in master? Could we merge master instead so that both develop and master are aligned?

JoshSinghONS:
> Aren't these different commits to the ones in master? Could we merge master instead so that both develop and master are aligned?

These are cherry picked from master. Would it not yield the same result?

Saves me discarding all the different changes between master and develop as well as well as trying to fix the numerous merge conflicts.

rafahop:
Well kind of, the code will match but we will have duplicated commits with half of them not present in develop which is not right. It seems like a simple change so the conflicts should be quite easy to fix?

JoshSinghONS:
> Well kind of, the code will match but we will have duplicated commits with half of them not present in develop which is not right. It seems like a simple change so the conflicts should be quite easy to fix?

https://github.com/ONSdigital/florence/pull/891

PR #896, opened 2022-08-18
bugfix session ending

geoffmrb:
### What

when logging into to florence, user is prompted with not being able to sign in

### How to review

review by checking the above 

### Who can review

anyone


joegasewicz (changes requested):
Blocking until i have seen this demo'd locally

PR #907, opened 2022-09-27
Feature/strengthens dataset type check

patrickjfl:
### What

Strengthens dataset type check to ensure that the dataset ID being used is present in Cantabular metadata server. This will further restrict the datasets that can be redirected to the 'Edit metadata (Cantabular)' screen and therefore should further protect the CMD journey for Team B to use for development purposes.

### How to review

Read the code, run the tests, for manual test instructions contact me for assistance in setting up a local environment.

### Who can review

Anyone but me or Valentina.

### DIFF
--- a/src/app/views/collections/details/CollectionDetailsController.jsx
+++ b/src/app/views/collections/details/CollectionDetailsController.jsx
@@ -594,7 +594,12 @@ export class CollectionDetailsController extends Component {
             const datasetID = pageID.split("/")[0].trim();
             const response = await datasets.get(datasetID);
             const type = response.next.type;
-            this.setState({ isCantabularDataset: type === "cantabular_flexible_table" || type === "cantabular_table" });
+            try {
+                await datasets.getCantabularMetadata(datasetID, "en");
+                this.setState({ isCantabularDataset: type === "cantabular_flexible_table" || type === "cantabular_table" });
+            } catch {
+                console.log("Dataset ID not present in Cantabular metadata server");
+            }
         }
     };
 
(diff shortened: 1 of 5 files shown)


src/app/views/collections/details/CollectionDetailsController.jsx
```diff
@@ -594,7 +594,12 @@ export class CollectionDetailsController extends Component {
             const datasetID = pageID.split("/")[0].trim();
             const response = await datasets.get(datasetID);
             const type = response.next.type;
-            this.setState({ isCantabularDataset: type === "cantabular_flexible_table" || type === "cantabular_table" });
+            try {
+                await datasets.getCantabularMetadata(datasetID, "en");
+                this.setState({ isCantabularDataset: type === "cantabular_flexible_table" || type === "cantabular_table" });
+            } catch {
+                console.log("Dataset ID not present in Cantabular metadata server");
```
gilesburdett, line 601: If Florence has a way to display error messages to the user, whether it's a banner or popup, by calling `setState()` with the error message, this would be much more helpful to the user as most people don't have the development console open and will simply think Florence is broken when it appears to do nothing.

patrickjfl: Definitely agree in general, but this a sort of special case because this screen is shared between the Cantabular dataset and CMD journeys. If the dataset ID is not in the Cantabular metadata server it's not necessarily an error for the CMD journey, so it would be strange to raise an error here, or even a warning. So we thought, useful information in the console but not necessarily beyond that.

gilesburdett (approved):

PR #918, opened 2022-10-11
Feature/1183 census branding

patrickjfl:
### What

Adds census branding radio button to Cantabular Edit metadata form.

Also refactors handling of canonical and subtopics to pass and return only topic IDs, rather than topic objects.

### How to review

Run tests, run through the Cantabular metadata journey, check that census branding persists and can be toggled.

### Who can review

Anyone


rafahop (dismissed):
Looks good to me

stmuk (approved):

PR #919, opened 2022-10-12
refactors dimension integration

patrickjfl:
### What

Refactors setting of "dimensions" metadata state to ensure Cantabular dimensions are only integrated before saving the updated metadata (after which point all metadata should come from the dataset-api, according to current requirements). Also adds error handling and fixes canonicalTopic type in view file.

### How to review

Look at the changes, run the journey if possible.

### Who can review

Anyone.


vmilitaru (approved):
👍

PR #922, opened 2022-10-20
Send language to data endpoint

rafahop:
### What

Fix an error when reading Welsh dataset information if no English equivalent dataset exists

### How to review

- Edit or create a Welsh dataset
- Download options > Add spreadsheet
- An error should not appear and the new period label show in the new element

### Who can review

Anyone


vmilitaru (approved):
👍

PR #928, opened 2022-11-01
reverts Cantabular dataset title to just label field

patrickjfl:
### What

Simplifies title field for Cantabular datasets, using only label rather than concatenating name (i.e. dataset code) + label. As requested by outputs.

### How to review

Check tests pass, sense check.

### Who can review

Anyone


vmilitaru (approved):
LGTM 👍

PR #935, opened 2022-11-25
Adds update cantabular metadata feature

vmilitaru:
### What

Adds `update cantabular metadata feature` to the Edit Metadata (Cantabular) form, displayed when the user first lands on the screen or on page refresh. The feature is accessible to the user only if the collection is `inProgress`. ([trello ticket](https://trello.com/c/bQI8Kqju/1299-update-metadata-feature)).

### How to review

Run the code, go through Cantabular metadata journey and at the end only click `Save`  to keep the collection `inProgress`. To test the  feature you can either go and manually change some of the dataset fields in mongo(from the mongo `datasets` collection you can manually change `title`,`description`,`keywords` etc and from the `instances` mongo collection you can manually change from the `dimensions` field `label`,`description`,`quality_statement_text` etc) or you can go and manually update/hard-code the `dp-cantabular-metadata-extractor-api` response.

Check the below example where the contact name has been changed in mongo.
When you first land on the page you get a pop out letting you know that there are some differences between the saved cantabular metadata to the existing dataset and the latest cantabular metadata returned by the `dp-cantabular-metadata-extractor-api`.
The user has 2 options:
1. Check the changes: 
 
Screenshot 2022-11-25 at 13 19 59

And this is the current response of the `dp-cantabular-metadata-extractor-api`:

Screenshot 2022-11-25 at 13 20 20

At this stage the user can choose to `Revert` or `Save` the incoming change. If the user decides to `Revert` then the contact name saved to the dataset in mongo gets displayed.

Screenshot 2022-11-25 at 13 20 33

Else if the user accepts the incoming change the new cantabular contact name gets saved to the dataset in mongo.

Screenshot 2022-11-25 at 13 42 19

2. Reject the changes so that everything stays as it is.

Screenshot 2022-11-25 at 13 36 52

 Now the displayed data is the dataset cantabular metadata saved in mongo.

Screenshot 2022-11-25 at 13 38 14

### Who can review

Anyone


patrickjfl (approved):
LGTM, nice work 👍

PR #949, opened 2023-01-18
Publishing queue performance improvements

jondewijones:
### What

- Delay call to collectionDetails end point and only call if a user requests this information. Originally calling to get all the details was slow particular with large collections - https://trello.com/c/hRMo256Y/549-publishing-queue-load-performance-issues-s2
- Remove unused "Remove from collection" button and replace with content url. It done nothing anyways - https://trello.com/c/qKvjO0xR/1401-remove-remove-from-this-publish-unused-button
- Tidy up some bits or code and remove commented out code

### How to review

Note:
Some of the code isn't as pretty as it could be but it was decided for consistency sake and not to rewrite large parts of this as we have no test coverage to check any re-write still works and this is in the old part of Florence which is maintenance/EOL support. 

### Who can review

Anyone 👋


src/legacy/js/functions/_viewPublishDetails.js
```diff
@@ -7,58 +9,31 @@ function viewPublishDetails(collections) {
...
+                pageType: 'manual',
+                showFilesButton: true
+            })
+        } else {
+            result.collectionDetails.push({
+                id: collection.id, 
+                name: collection.name,
+                showFilesButton: true
+            })
+        }
```
rafahop, line 27: Since the only difference seems to be the pageType, would something like this be easier to read/maintain?

```suggestion
        result.collectionDetails.push({
            id: collection.id, 
            name: collection.name,
            pageType: collection.date === manual ? "manual" : ""
            showFilesButton: true
       })
```

rafahop (approved):
Looks good to me 🙂

src/legacy/js/functions/_viewPublishDetails.js
jondewijones: i like it 👍

PR #951, opened 2023-01-25
Feature/hide users tabs for non admins

lindenmckenzie:
### What

Hidden 'Users and access' tab in florence and legacy from non-admins. 
Fixed a redirect for editors switching between legacy and non-legacy

### How to review

Initial issue was editors and viewers could see 'users' tab - this is now not rendered for them.
The subsequent issue was when changing back and forth between legacy (in this case, publishing-queue), the Router redirects 'editors' to the 'collections' page, which looks to be incorrect as once in place in the React app it allows access just fine. Logic looked incorrect on closer inspection, so I've corrected - happy to take it out if this is expected behaviour. 

### Who can review

Someone with Florence knowledge.

### DIFF
--- a/src/app/components/layout/Layout.test.js
+++ b/src/app/components/layout/Layout.test.js
@@ -51,6 +51,7 @@ jest.mock("../../utilities/api-clients/user", () => {
     class user {}
     user.setUserState = jest.fn();
     user.getOldUserType = jest.fn();
+    user.getUserRole = jest.fn();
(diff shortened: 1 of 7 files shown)


nshumoogum (approved):
LGTM, Im not an expert in this product. So could do with a frontend developer in Team RTT to have a look at this PR

jondewijones (dismissed):

PR #959, opened 2023-02-17
setting cookie for collection when clicked on preview button.

yomiodeleye:
### What

Previewing an approved collection in Florence only works for the first previewed collection per session.

### How to review

Create/ensure two different collections ('A' &' B') with some edited content are in an ‘approved’ state (ie. they appear in the publishing queue).
Sign out of Florence then back in again (clear session)
Go to "Publishing queue" -> [date]/[manual collection] -> [collection 'A'] -> "Preview" and observe that the edited content CAN be previewed (expected collection changes are visible)
Go to "Publishing queue" -> [date]/[manual collection] -> [collection ‘B'] -> "Preview" and observe that the content changed in collection ‘B’ CAN NOT be previewed (only the content in the published master is viewable). However, note that the content modified in collection 'A’ is still being overlaid and shows the modified content rather than the master as it should do.
Sign out of Florence then back in again (clear session)
Go to "Publishing queue" -> [date]/[manual collection] -> [collection 'B'] -> "Preview" and observe that the edited content nowCAN be previewed (expected collection changes are visible) as it is now the first collection previewed in this session.

### Who can review

FD

### DIFF
--- a/src/legacy/js/functions/_viewPublishDetails.js
+++ b/src/legacy/js/functions/_viewPublishDetails.js
@@ -1,132 +1,152 @@
 function viewPublishDetails(collections) {
+  var manual = "[manual collection]";
+  var result = {
+    date: Florence.collectionToPublish.publishDate,
+    subtitle: "",
+    collectionDetails: [],
+    pendingDeletes: [],
+  };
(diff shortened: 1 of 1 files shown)


jondewijones (approved):

vmilitaru (approved):
👍

PR #967, opened 2023-03-02
Fix/renaming collection

JoshSinghONS:
### What

Added to validation to ensure a collection can't be created if a collection ID exists with the same prefix.

### How to review

LGTM.

### Who can review

Anyone.


src/app/views/collections/validation/collectionValidation.js
```diff
@@ -33,6 +33,13 @@ export default class collectionValidation {
             };
         }
 
+        if (collections && collections.some(c => c.id.split("-")[0] === name.replace(/\s+/g, ""))) {
```
rafahop, line 36: Can we add test coverage for `name.replace(/\s+/g, "")` please

src/app/views/collections/validation/collectionValidation.js
```diff
@@ -33,6 +33,13 @@ export default class collectionValidation {
...
+            response = {
+                isValid: false,
+                errorMsg: "A collection with this ID already exists. Please choose a different name.",
```
rafahop, line 39: Not sure if we should add this error to `errCodes` like for unique name?

rafahop (commented):
This new validation is only needed when a collection has been renamed. Only in that case the name and id won't match. 
However, if the collection has not been renamed, I think the existing unique name error will stop showing and this will show instead as it overrides the `response` object. Is this the expected behaviour? Could we add unit test for that scenario?

src/app/views/collections/validation/collectionValidation.js
JoshSinghONS: > This new validation is only needed when a collection has been renamed. Only in that case the name and id won't match. However, if the collection has not been renamed, I think the existing unique name error will stop showing and this will show instead as it overrides the `response` object. Is this the expected behaviour? Could we add unit test for that scenario?

Hello, 

Just so I understand the scenario you are suggesting because it took me a few passes to grasp it lol

If a collection is called alpha thus id: alpha and name: alpha. 

Then a second collection tries to be created called alpha.

Is the problem that the UNIQ_NAME_ERROR will never show and it will default to the "A collection with this ID already exists. Please choose a different name." error message? 

If that is your thinking then to be honest I don't know what the desired behaviour is in that instance and will need to check with the PO.

rafahop: Yes that is the scenario I was thinking about. Please check with the PO and if we are happy with that result, I think the existing name check is dead code and should be removed

rafahop: We should also add an unit test for that scenario. I notice the examples in the tests never have an id and name matching (and this is what most collection will have)

JoshSinghONS: Would it be dead code?

I mean what about a scenario where

A collection called alpha gets created thus id: alpha-123 and name: alpha

Then a second collection gets created called alphaTest thus id: alphaTest-123 and name: alphaTest

Then alphaTest gets renamed to alpha. Removing the existing name check would allow for that and cause issues, no?

rafahop: Yes, I think you are right there! 👍

src/app/views/collections/validation/collectionValidation.js
```diff
@@ -26,7 +26,10 @@ export default class collectionValidation {
             };
         }
 
-        if (collections && (collections.some(c => c.name === name.trim()) || collections.some(c => c.id.split("-")[0] === name.replace(/\s+/g, "")))) {
+        name = name.trim()
+        name = name.replace(/\s+/g, "")
```
rafahop, line 30: Wouldn't the `replace` function make the `trim` unnecessary? 
Also, I'd suggest using a new variable instead of reusing `name` - it is harmless _at the moment_ as we don't use `name` after this but it is good practice not to reassign function parameters

rafahop: Actually, no. If we want to keep the current behaviour, we need both replace and trim, but in different variables. We check the name against a trimmed version of the current name but the id against the whitespace-free version of the name. If we replace all the spaces, the name check would fail. Do the unit tests pass with this code?

JoshSinghONS: They do.

rafahop (commented):
Code looks ok to me 👍 
However when running this locally, I get changes in package-lock.json, shouldn't these be checked in?

rafahop (approved):

PR #979, opened 2023-04-18
Bug Fix for conditional due to id being compared to a string which was not converted to lower case.

JoshSinghONS:
### What

Title

### How to review

LGTM.

### Who can review

Anyone.

### DIFF
--- a/src/app/views/collections/validation/collectionValidation.js
+++ b/src/app/views/collections/validation/collectionValidation.js
@@ -27,7 +27,7 @@ export default class collectionValidation {
         }
 
         let nameTrimmed = name.trim();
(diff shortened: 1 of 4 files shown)


src/app/views/collections/validation/collectionValidation.js
```diff
@@ -27,7 +27,7 @@ export default class collectionValidation {
         }
 
         let nameTrimmed = name.trim();
-        let nameReplaced = name.replace(/\s+/g, "");
+        let nameReplaced = name.replace(/\s+/g, "").toLowerCase();
```
rafahop, line 30: Any chance we could have test coverage for this please?

Maybe also worth renaming the variable?

rafahop (approved):

PR #1009, opened 2024-02-21
WIP

JoshSinghONS:
### What

Proposed changes to fix bug where when secauth is enabled instead of showing the last user who edited a page's email it instead shows their UUID.

More information in this [ticket](https://trello.com/c/LyLcbO0E/1967-secauth-florence-shows-uuid-not-email-for-edits-and-more)

### How to review

Required repos dp-permissions-api, dp-identity-api, zebedee, dp-api-router, florence.
Make sure local florence points to this branch.
Run the secauth stack found in [dp-compose](https://github.com/ONSdigital/dp-compose/tree/main/v2/stacks/auth)
Read my comments and tell me your thoughts.

### Who can review

Anyone.

### DIFF
--- a/src/app/utilities/api-clients/user.js
+++ b/src/app/utilities/api-clients/user.js
@@ -94,8 +94,8 @@ export default class user {
     static createNewUser(body) {
         return http.post(`/users`, body);
     }
-    // TODO: new auth work
-    static getUser(id) {
+
+    static getUserEmail(id) {
         return http.get(`/users/${id}`);
     }
 
--- a/src/app/views/collections/details/CollectionDetailsController.jsx
+++ b/src/app/views/collections/details/CollectionDetailsController.jsx
@@ -176,6 +176,7 @@ export class CollectionDetailsController extends Component {
                     return;
                 }
                 const mappedCollection = collectionMapper.collectionResponseToState(collection, this.props.groups);
+                // Have some stuff going on here?
                 const collectionWithPages = collectionMapper.pagesToCollectionState(mappedCollection);
 
                 // If we have no data in state yet for the collection then use this opportunity to add it.
@@ -648,6 +649,7 @@ export class CollectionDetailsController extends Component {
                 isLoadingDetails={this.state.isFetchingCollectionDetails}
                 isCancellingDelete={this.state.isCancellingDelete}
                 isApprovingCollection={this.props.isUpdating}
+                isNewSignIn={this.props.isNewSignIn}
             />
         );
     }
--- a/src/app/views/collections/mapper/collectionMapper.js
+++ b/src/app/views/collections/mapper/collectionMapper.js
@@ -100,6 +100,15 @@ export default class collectionMapper {
                             uri: page.uri,
                             type: page.type,
                         };
+                        // if (this.props.isNewSignIn) {
+                        //     user.getUserEmail(updatedPage.lastEdit.email)
+                        //         .then(response => {
+                        //             return response;
+                        //         })
+                        //         .then(data => {
+                        //             updatedPage.lastEdit.email = data.email;
+                        //         });
+                        // }
                     } catch (error) {
                         log.event("Error mapping a page to Florence's state", log.error(error));
                         console.error("Error mapping a page to Florence's state", error);
--- a/src/app/views/collections/details/CollectionDetails.jsx
+++ b/src/app/views/collections/details/CollectionDetails.jsx
@@ -8,6 +8,7 @@ import url from "../../../utilities/url";
 import log from "../../../utilities/logging/log";
 import Page from "../../../components/page/Page";
 import date from "../../../utilities/date";
+import user from "../../../utilities/api-clients/user.js";
 
 export const pagePropTypes = {
     lastEdit: PropTypes.shape({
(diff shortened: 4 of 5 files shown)


src/app/views/collections/details/CollectionDetailsController.jsx
```diff
@@ -176,6 +176,7 @@ export class CollectionDetailsController extends Component {
                     return;
                 }
                 const mappedCollection = collectionMapper.collectionResponseToState(collection, this.props.groups);
+                // Have some stuff going on here?
```
JoshSinghONS, line 179: I think this may be a potential place to put code making requests to /users however here there would need to be some array manipulation and just isn't as easy as it is in details.

src/app/views/collections/details/CollectionDetailsController.jsx
```diff
@@ -176,6 +176,7 @@ export class CollectionDetailsController extends Component {
...
                 const collectionWithPages = collectionMapper.pagesToCollectionState(mappedCollection);
```
JoshSinghONS, line 181: I think this may be a potential place to put code making requests to /users however here there would need to be some array manipulation and just isn't as easy as it is in details. Operation would occur after mapping but before being passed to details for render

src/app/views/collections/mapper/collectionMapper.js
```diff
@@ -100,6 +100,15 @@ export default class collectionMapper {
...
                         };
+                        // if (this.props.isNewSignIn) {
+                        //     user.getUserEmail(updatedPage.lastEdit.email)
+                        //         .then(response => {
+                        //             return response;
+                        //         })
+                        //         .then(data => {
+                        //             updatedPage.lastEdit.email = data.email;
+                        //         });
+                        // }
```
JoshSinghONS, line 111: Thought about putting it here too with the idea of making the users/id requests a part of the mapping process however the way the application is structured I do not think its possible to have access to the environment variables i.e. isNewSignIn in this case.

src/app/views/collections/details/CollectionDetailsController.test.js
```diff
@@ -734,4 +740,56 @@ describe("When the component mounts with a collection id", () => {
...
+                approvalStatus: "NOT_STARTED",
+                events: [
+                    {
+                        date: "2024-02-12T10:43:10.803Z",
+                        type: "CREATED",
+                        email: "test.user@email.com",
+                    },
+                ],
+            })
+        );
```
lindenmckenzie, line 784: Not a review but this isn't working as you're anticipating - the mockcalls from earlier up are actually getting fired and it's not returning this when it gets the collection. 

Might want to try resetting the mocks between test runs or running them synchronously (I'm not sure if they're async atm) or using different mock data. 

It doesn't actually *solve* the problem per se, but it does get you further forward. Try more logs to see where the code is getting to and what's actually getting triggered. At the moment the getUserEmail function is *not* being called so the test is reporting correctly.

lindenmckenzie (commented):
Some comments to unblock you.

src/app/utilities/api-clients/user.js
```diff
@@ -94,8 +94,8 @@ export default class user {
     static createNewUser(body) {
         return http.post(`/users`, body);
     }
-    // TODO: new auth work
-    static getUser(id) {
+
+    static getUserEmail(id) {
```
jondewijones, line 98: I might be wrong but doesn't this still return the user object, rather than the email? The previous name might be more descriptive

src/app/views/collections/details/CollectionDetailsController.jsx
```diff
@@ -305,7 +306,27 @@ export class CollectionDetailsController extends Component {
...
     };
 
-    handleCollectionPageClick = uri => {
+    handleCollectionPageClick = async (page, state) => {
+        const uri = page.uri;
+        if (this.props.isNewSignIn) {
+            const updateEmailInActiveCollection = {
+                ...this.props.activeCollection,
+            };
+            await user.getUserEmail(page.lastEdit.email).then(response => {
```
jondewijones, line 315: Is it worth considering what we do if this request fails/errors? And might be worth considering what we show the user while we wait

jondewijones: We'd probably want to add a log event as well as console log. e.g.`log.event("error getting user email address", error.log(error))` so that we send the error server side and is captured in Kibana.

Locally it will be fairly quick. Over a slow network and the VPN (which our users have to use) it might not be so quick. It's worth simulating slow network when testing such things. Dev tools/Inspect > Network > Throttling (slow 3g).

As with most things in newer parts of Florence i'd suggest we have a loading state, a loaded state and an error state. Something like:

Loading:
`Last edit: loading...`

or
`Last edit: loading... at (Mon 22 Jan 2024 - 15:20:17)`

Loaded: 
`Last edit: user@user.com (Mon 22 Jan 2024 - 15:20:17)`

Failed:
`Last edit: {UUID} at(Mon 22 Jan 2024 - 15:20:17)`

or even something to say it's failed. I don't think we need anything crazy and would be against designing something over the top. 

Happy for it to be run past UR first so it can be a new ticket. A fix the bug fix the bug ticket ;)

jondewijones (approved):

PR #1020, opened 2024-05-07
Enable teams screen for publishing role

jondewijones:
### What

Enable teams screen for publishing role

### How to review

Log in as publisher role user and check can access "Preview teams" screen, screens work/match UX designs
Log in as admin role user check "Preview teams" screen works as before

### Who can review

Anyone


dionisislolas:
In the screenshot at the top of the ticket at the right screenshot we can see there are no 'Save changes' and 'Cancel' buttons when we are previewing the members of the group. But when I login as a publisher I can see them both. Is this the expected behaviour?

dionisislolas (approved):

PR #1021, opened 2024-05-09
Allow read-only user screens to editors

lindenmckenzie:
### What

Allowed editors to access Users and Access screens in a read-only version. 

### How to review

Switch on EnableNewSignIn - port forward 23200 to sandbox api router. 
Login as editor - check looks good
Login as admin - check looks good. 

### Who can review

Frontend react dev.


jondewijones (approved):
LGTM

PR #1037, opened 2024-07-15
Feature/search api for release calendar entry

StRhys:
### What

Previously florence was using babbage to get release calendar entries, but due to decommissioning the release calendar in babbage this needed to be switched to use the search api [instead](https://jira.ons.gov.uk/browse/DIS-1106). 

### How to review

- Grab this branch
- Run the search stack
- Run florence
- Login
- Check "Calendar entry schedule"
- Click "Select a calendar entry"
and if all is working, you should see release calendar entries

image
image

You can also check it against sandbox. By changing `APIRouterURL` in config to `https://api.dp.aws.onsdigital.uk` then doing `make debug` again while still logged in and refreshing. The calendar entries should now pull from sandbox.

image

* these entries won't directly match current sandbox florence, as currently babbage pulls all releases that are nott published and no cancelled, while the search api returns the not published, not cancelled _AND_ from date/time now.

### Who can review

Florence dev

### DIFF
--- a/service/handlers.go
+++ b/service/handlers.go
@@ -13,6 +13,8 @@ import (
 	"github.com/ONSdigital/florence/config"
 	"github.com/ONSdigital/log.go/log"
 	"github.com/gorilla/mux"
+
+	dphandlers "github.com/ONSdigital/dp-net/v2/handlers"
 )
 
 // generated files constants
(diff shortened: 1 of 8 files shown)


service/service.go
```diff
@@ -179,6 +184,7 @@ func (svc *Service) createRouter(ctx context.Context, cfg *config.Config) (route
 
 	// API and Frontend Routers
 	router.Handle("/api/{uri:.*}", apiRouterProxy)
+	router.HandleFunc("/releasecalendar/data", ReleaseData(sc))
```
lindenmckenzie, line 187: Couldn't we just point the request at the apiRouterProxy? Or does that not work?

src/app/utilities/api-clients/releases.js
```diff
@@ -5,19 +5,17 @@ export default class releases {
...
         }
 
         if (size) {
-            parameters.push("size=" + size);
+            parameters.push("limit=" + size);
         }
-
-        const URL = "/releasecalendar/data?view=upcoming&" + parameters.join("&");
-
+        const URL = "/releasecalendar/data?release-type=type-upcoming&" + parameters.join("&");
```
lindenmckenzie, line 18: I think retaining this URL may cause confusion.

service/handlers.go
```diff
@@ -188,3 +190,27 @@ func DeleteHttpCookie() http.HandlerFunc {
...
+			log.Event(ctx, "error marshalling releases", log.ERROR, log.Error(err))
+			return
+		}
+
+		if _, err = w.Write(data); err != nil {
+			log.Event(ctx, "error writing response", log.ERROR, log.Error(err))
+			return
+		}
+	})
+}
```
lindenmckenzie, line 216: Don't think I can see any tests for this.

lindenmckenzie (changes requested):
A few thoughts - probably needs some further discussion, but this is a great start.

lindenmckenzie (changes requested):

service/interfaces.go
```diff
@@ -5,6 +5,7 @@ import (
 	"net/http"
 
 	"github.com/ONSdigital/dp-api-clients-go/v2/health"
+
 	"github.com/ONSdigital/dp-healthcheck/healthcheck"
 	"github.com/ONSdigital/florence/config"
```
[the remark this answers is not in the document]
StRhys: health and healthcheck were getting a bit close for my liking

lindenmckenzie (approved):

PR #1053, opened 2024-09-19
Account for topics with no subtopics

lindenmckenzie:
### What

Add check for if no subtopics (introduced by new parent topics without subtopics)

### How to review

Check code good, tests pass. 

Optionally:

Run Florence locally with SecAuth on, port forward to api router in sandbox. 
Create collection, edit bulletin page (or other with tags). Observe tagging now works appropriately. 

### Who can review

Florence dev.


dionisislolas (approved):
LGTM

PR #1086, opened 2024-12-19
Release/1.83.0

mr-nick17:
### What

- ✅ [Remove all old auth from Florence](https://jira.ons.gov.uk/browse/DIS-2086)

### How to review

Sense check
CI checks pass
Nothing's there that shouldn't be

### Who can review

!me


jessjenkins (approved):
Makes sense, checks pass and nothing seems to be there that shouldn't 👍

PR #1093, opened 2025-01-29
Feature/replace client side renewal

StRhys:
### What

[This](https://jira.ons.gov.uk/browse/DIS-2085) was to replace client side renewal with the new [dis-authorisation-client-js](https://github.com/ONSdigital/dis-authorisation-client-js).

This covers both react and legacy

### How to review

console logs are on to see what is happening.

Run florence

- port forward to the api router - `dp ssh sandbox publishing 2 -p 23200:localhost:10800`
- `make dev` in florence

### Who can review

Not me.


mr-nick17 (approved):
LGTM

PR #1114, opened 2025-04-30
Feature/remove topic proxy

lindenmckenzie:
### What

- Removed Topic API proxy in favour of API router proxy
- Removed redundant Interactives action type (no longer used or exists)
- Refactor Identity API calls to use generic proxy constants to match Topic API

There is duplication here between Legacy and React - this is to follow the existing conventions of no cross-pollution between the two. 

In writing this PR I did consider abstracting the base path to config as well but thought it was not worth the effort right now. I also considered that we could just wrap all the Florence requests in the Versioned Paths but am uncertain if this is possible at the current time. If we deprecated all the proxies and find that we can do this then we can refactor at that point. 
 
### How to review

Check looks ok and seems sensible approach. 

Optionally, to test:

- Run florence (make watch-src & make debug)
- Port forward to sandbox dp-api-router
- Check requests to `/topics` are now 404ing
- To test the topic request changes:
   - make a collection
   - edit a page
   - see that the topic selector in 'tags' populates and no console errors
   - save / check it re-appears etc.
- To test the identity request changes:
  - login
  - view all users
  - edit a user
  - view all groups
  - edit a group 

### Who can review

Not me.

### DIFF
--- a/src/app/utilities/api-clients/constants.js
+++ b/src/app/utilities/api-clients/constants.js
@@ -1,0 +1,7 @@
+export const API_PROXY = {
+    BASE_PATH: "/api",
+    VERSION: window.getEnv().apiRouterVersion,
+    get VERSIONED_PATH() {
+        return `${this.BASE_PATH}/${this.VERSION}`;
+    },
+};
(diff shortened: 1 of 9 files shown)


src/app/utilities/api-clients/constants.js
```diff
@@ -0,0 +1,7 @@
+export const proxy = {
```
mr-nick17, line 1: how about 
```suggestion
export const api = {
```

mr-nick17 (approved):

PR #1116, opened 2025-05-01
Remove files proxy

lindenmckenzie:
### What

Removed the Files API proxy

### How to review

Check looks ok and tests pass. 

I don't know how to test it but it's also switched off in all environments and I believe last time it was switched on it didn't work. Questionable whether we should keep it there but I'm choosing to leave it there for now. 

### Who can review

Not me.


Oluwawunmi (approved):
LGTM

PR #1122, opened 2025-05-14
Add missing file imports

lindenmckenzie:
### What

Add missing imports from upload / download service deprecation

### How to review

Run florence (make watch-src & make debug) 
Port forward api-router to sandbox
Login to Florence
Go to Uploads screen and select a dataset, it will now load. 

### Who can review

Not me.


franmoore05 (approved):
LGTM

PR #1127, opened 2025-05-16
fix typos in legacy js causing runtime failures

jessjenkins:
### What

Fix two JS typos causing runtime failures due to missing back-quotes

### How to review

Tests pass and changes look valid

### Who can review

Anyone


cookel2 (approved):
LGTM

PR #1139, opened 2025-06-09
Remove login form

lindenmckenzie:
### What

Remove LoginForm as no longer used. 

### How to review

Check looks ok - haven't missed any references

### Who can review

Not me.


cookel2 (approved):
LGTM

PR #1145, opened 2025-06-16
Fix/brace expansion

lindenmckenzie:
### What

- Upgraded x/net
- Ignored x/net vuln as latest version 0.41 is affected (can see fix is in pipeline)
- Overridden brace-expansion to receive patch version

### How to review

Check looks ok, tests pass.

### Who can review

Not me.


redhug1 (approved):
LGTM

PR #1149, opened 2025-06-25
Feature/add migration field non series types

mr-nick17:
### What

Added migration field to types: `taxonomy_landing_page` and `release` - aka ✅ [Add migrationLink field to priority editorial non-series types in Florence](https://jira.ons.gov.uk/browse/DIS-3283)
Also fixed an annoying bug when editing a taxonomy landing page that causes the `Related highlights` accordion to always be open and unclosable 

### How to review

Sense check
_Optionally_
- Pull this branch
- make debug ENABLE_MIGRATION_FIELD=true
   - you may need to rebuild node_modules
- Port forward to the api router in sandbox (publishing mount)
- Test various scenarios of your choosing

### Who can review

Frontend dev

### DIFF
--- a/src/legacy/js/constants/sweetAlerts.js
+++ b/src/legacy/js/constants/sweetAlerts.js
@@ -1,0 +1,1 @@
+const MIGRATION_FIELD_VALIDATION_FAILURE = ["Cannot save this page", "Migration path must be a relative path starting with '/'"];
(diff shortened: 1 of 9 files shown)


src/legacy/js/constants/migration.js
```diff
@@ -0,0 +1 @@
+const MIGRATION_FIELD_VALIDATION = ["Cannot save this page", "Migration path must be a relative path starting with '/'"];
```
lindenmckenzie, line 1:
```suggestion
const MIGRATION_FIELD_VALIDATION_FAILURE = ["Cannot save this page", "Migration path must be a relative path starting with '/'"];
```

lindenmckenzie (changes requested):
This is good - please consider naming your const and it's corresponding file more clearly.

PR #1190, opened 2026-02-12
Make navbar more consistent

lindenmckenzie:
### What

Made the nav bar more consistent between the two parts of Florence.

AdminOrEditor should now be able to see:
- Datasets
- Publishing Queue
- Users and Access
- Preview Teams

Admin only should see:
- Security

All should see:
- Collections
- Logout

### How to review

Run it, login (may involve complicated setup with the legacy-core-publishing stack) and see you can see the right things. 

### Who can review

Not me.


cookel2 (approved):
Tested successfully

PR #1196, opened 2026-03-04
Feature/remove collection policy on delete

dhee-tree:
### What

- [Delete collection permissions policy on collection publish or deletion](https://officefornationalstatistics.atlassian.net/browse/DIS-4555)
- [Fix error thrown when opening edit collection screen if no viewer groups](https://officefornationalstatistics.atlassian.net/browse/DIS-4557)
- [Fix errors when editing collections](https://officefornationalstatistics.atlassian.net/browse/DIS-4556)

### How to review

- Sense check
- Run branch with `legacy-core-publishing`
  - SSH to identity API in sandbox and use `make up-with-seed`
- Login to http://localhost:8081/florence/login with sandbox creds
- Review policies collection in permissions database
- Test collection without a team 
  - Create collection - policy should not be created
  - Edit screen should have no UI errors (404 getting a policy is expected in devtool)
    - In edit screen add a team or two
      - No errors, policy should be created with team(s)
      - Removing a team or all teams should remove that team or delete policy
- Test collection with a team
  - On creating a collection with a team a policy should be created
- Deleting a collection with a team policy should be deleted

### Who can review

!Me


src/app/config/thunks.js
```diff
@@ -440,12 +440,64 @@ export const updatePolicyRequest = (id, body) => async dispatch => {
...
+    } catch (error) {
+        if (error.status === 404) {
+            dispatch(actions.deletePolicySuccess(null));
+            return;
+        }
+        dispatch(actions.deletePolicyFailure());
+        switch (error.status) {
+            case 401: {
+                break;
+            }
```
lindenmckenzie: Why the break if it's a 401?

src/app/config/thunks.js
```diff
@@ -440,12 +440,64 @@ export const updatePolicyRequest = (id, body) => async dispatch => {
...
+            case 400: {
+                const notification = {
+                    type: "warning",
+                    message: "There was an error deleting the collection policy. Please try again.",
```
lindenmckenzie: Are they able to try again though?

src/app/config/thunks.js
```diff
@@ -440,12 +440,64 @@ export const updatePolicyRequest = (id, body) => async dispatch => {
...
+        const result = await collections.deletePolicy(id);
+        dispatch(actions.deletePolicySuccess(result));
+    } catch (error) {
+        if (error.status === 404) {
+            dispatch(actions.deletePolicySuccess(null));
+            return;
+        }
+        dispatch(actions.deletePolicyFailure());
+        switch (error.status) {
+            case 401: {
```
lindenmckenzie: better to order these by number order.

src/app/config/thunks.js
```diff
@@ -440,12 +440,64 @@ export const updatePolicyRequest = (id, body) => async dispatch => {
...
 export const loadPolicyRequest = (id, body) => async dispatch => {
     dispatch(actions.loadPolicyProgress());
     try {
         const result = await collections.getPolicy(id, body);
         dispatch(actions.loadPolicySuccess(result));
     } catch (error) {
+        if (error.status === 404) {
+            dispatch(actions.loadPolicySuccess(null));
+            return;
+        }
```
lindenmckenzie, line 506: What if it fails to talk to the API or it gets a wrong address though? 

Either we need to call this anti-pattern in a comment or we should return the error and let the collection screen handler deal with this logic.

src/app/views/collections/details/CollectionDetailsController.jsx
```diff
@@ -215,6 +215,9 @@ export class CollectionDetailsController extends Component {
         collections
             .delete(collectionID)
             .then(async () => {
+                if (this.props.isEnablePermissionsAPI && this.props.activeCollection?.teams?.length > 0) {
+                    this.props.dispatch(deletePolicyRequest(collectionID));
```
lindenmckenzie, line 219: If this fails does it block deleting the collection?

src/app/views/collections/details/CollectionDetailsController.jsx
```diff
@@ -233,6 +236,9 @@ export class CollectionDetailsController extends Component {
 
     handleCollectionApproveClick = e => {
         this.props.dispatch(approveCollectionRequest(this.props.collectionID, `${this.props.rootPath}/collections`));
+        if (this.props.isEnablePermissionsAPI && this.props.activeCollection?.teams?.length > 0) {
+            this.props.dispatch(deletePolicyRequest(this.props.collectionID));
```
lindenmckenzie, line 240: Same question as above here - if we fail to delete the policy, does the approval still happen?

src/legacy/js/functions/_publish.js
```diff
@@ -53,3 +53,24 @@ function unlock(collectionId) {
...
+        attribute: "collection_id",
+        operator: "StringEquals",
+        values: [collectionId],
+      },
+    }),
+    error: function (response) {
+      handleApiError(response);
+    }
+  });
+}
```
lindenmckenzie: would put this in it's own dedicated function file, rather than bundling with "_publish"

lindenmckenzie (commented):
Some minor bits, more around the desired outcomes rather than the code - will now give it a test

src/app/config/thunks.js
dhee-tree: Added an error message

src/app/config/thunks.js
dhee-tree: Try again removed

src/app/config/thunks.js
dhee-tree: Done

src/app/views/collections/details/CollectionDetailsController.jsx
dhee-tree: No, it will not block collection deleting as collection is done before

src/app/views/collections/details/CollectionDetailsController.jsx
dhee-tree: yes, approval still happens

src/legacy/js/functions/_publish.js
dhee-tree: removed as per changes to ticket

lindenmckenzie (approved):

PR #1203, opened 2026-03-26
Add feature flagged system nav bar

lindenmckenzie:
### What

Added a feature flagged system nav bar to navigate between dissemination apps. Styling is largely copied from the internal service header from the service manual but marginally adapted to match Florence (e.g. full width).

It also adds a new front screen when you login instead of 'collections'

Switched off by default via `ENABLE_SYSTEM_NAV_BAR`

### How to review

Check looks ok. 

To test you can run with the `legacy-core-publishing` stack and switch it on / off. If testing you should test on and off and inside the react part of Florence and the legacy part of Florence.

### Who can review

Not me.


jondewijones (approved):
LGTM

PR #1214, opened 2026-04-09
Add read only migration field to dataset and dataset landing page

dhee-tree:
### What

[Add read only migration field to dataset and dataset_landing_page
](https://officefornationalstatistics.atlassian.net/browse/DIS-4553)

### How to review

- run florence locally with `make debug ENABLE_MIGRATION_FIELD=true`
- port forward the api-route, identity api and permissions api or use `legacy-core-publishing`
- login with sandbox creds
- create collection
- Edit a dataset i.e. `http://localhost:8081/economy/investmentspensionsandtrusts/datasets/ownershipofukshares/2014`
- Edit a dataset_landing_page i.e. `http://localhost:8081/economy/investmentspensionsandtrusts/datasets/ownershipofukshares`
- You should see the read only migration field

image

### Who can review

!Me

### DIFF
--- a/ci/scripts/lint.sh
+++ b/ci/scripts/lint.sh
@@ -3,6 +3,5 @@
 cwd=$(pwd)
 
 pushd $cwd/florence
-  go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6
   make lint
 popd
--- a/src/audit-allowlist.json
+++ b/src/audit-allowlist.json
@@ -19,6 +19,10 @@
         {
             "id": "CVE-2025-57352",
             "reason": "Requires upgrades to dp-table-builder-ui"
+        },
+        {
+            "id": "CVE-2026-25547",
+            "reason": "Patched in 5.0.1, but minimatch 3.x requires brace-expansion 1.x"
         }
     ]
 }
--- a/src/legacy/js/functions/_migration.js
+++ b/src/legacy/js/functions/_migration.js
@@ -2,12 +2,12 @@
  * manage migration link
  * @param templateData
  * @param data
+ * @param isReadOnlyMigrationPath
  */
 
-async function migration(templateData, data) {
+async function migration(templateData, data, isReadOnlyMigrationPath = false) {
     const migrationConfig = window.getEnv().enableMigrationField;
     if (migrationConfig) {
-        const isReadOnlyMigrationPath = data.type === 'dataset' || data.type === 'dataset_landing_page';
         templateData.readOnlyMigrationPath = isReadOnlyMigrationPath;
         let html = templates.migration(templateData)
         $('#migration').replaceWith(html);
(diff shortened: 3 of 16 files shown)


ci/scripts/lint.sh
```diff
@@ -3,6 +3,6 @@
 cwd=$(pwd)
 
 pushd $cwd/florence
-  go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6
+  go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4
   make lint
 popd
```
lindenmckenzie, line 7: I think there's a node-go-lint image now.

src/legacy/js/functions/_migration.js
```diff
@@ -7,13 +7,17 @@
 async function migration(templateData, data) {
     const migrationConfig = window.getEnv().enableMigrationField;
     if (migrationConfig) {
+        const isReadOnlyMigrationPath = data.type === 'dataset' || data.type === 'dataset_landing_page';
+        templateData.readOnlyMigrationPath = isReadOnlyMigrationPath;
```
lindenmckenzie: An alternate approach would be to pass the variable to the function - rather than letting the function decide which templates, let the template decide if it wants it to be readonly.

src/audit-allowlist.json
```diff
@@ -19,6 +19,10 @@
         {
             "id": "CVE-2025-57352",
             "reason": "Requires upgrades to dp-table-builder-ui"
+        },
+        {
+            "id": "CVE-2026-25547",
```
lindenmckenzie, line 24: I think brace-expansion 1.1.14 resolves this as per https://github.com/juliangruber/brace-expansion/compare/v1.1.13...v1.1.14

did you still get an audit failure with the override?

lindenmckenzie (changes requested):
I've read through the code, will have a look at the UI shortly.

src/legacy/js/functions/_migration.js
dhee-tree: Done, moved into template level

src/audit-allowlist.json
dhee-tree: yup, I still get the failure with 1.1.14

ci/scripts/lint.sh
dhee-tree: replaced with image

src/legacy/scss/partials/_workspace.scss
```diff
@@ -122,6 +122,19 @@
...
+                background-color: $quill-gray;
+                border: 3px solid $scarpa-flow;
+                color: $dusty-gray;
+                cursor: not-allowed;
+            }
+
+            #migration_link[readonly]:focus,
+            #migration_link[aria-readonly="true"]:focus {
+                border: 3px solid $scarpa-flow !important;
+            }
```
lindenmckenzie: Could we make this more generic? i.e. for all text areas with readonly

lindenmckenzie: Obviously check if there are clashes with anything else that is readonly.

dhee-tree: Done. Now generic for all textareas inside the workspace edit and will not clash with other readonly because it only applies to .workspace-edit

lindenmckenzie (approved):

PR #1220, opened 2026-04-20
Fix/automated collection list

lindenmckenzie:
### What

Automated collections aren't showing up in the list in florence with 'automated collection'
I've also marked a series of functions / template in legacy florence as deprecated.

### How to review

You can run florence with zebedee and change a collection to 'automated' via the content store - then see it in the list. This can be most easily done with the legacy-core-publishing stack

### Who can review

Not me.

### DIFF
--- a/src/app/views/collections/mapper/collectionMapper.js
+++ b/src/app/views/collections/mapper/collectionMapper.js
@@ -245,9 +245,7 @@ export default class collectionMapper {
 
         if (collection.publishDate) {
             return date.format(collection.publishDate, "ddd, dd/mm/yyyy h:MMTT");
-        }
-
-        if (!collection.publishDate) {
+        } else {
             // This check should be a switch with manual but this
             // approach preserves the existing logic better in this
             // instance.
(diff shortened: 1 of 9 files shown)


src/app/views/collections/mapper/collectionMapper.js
```diff
@@ -248,6 +248,13 @@ export default class collectionMapper {
         }
 
         if (!collection.publishDate) {
+            // This check should be a switch with manual but this
```
cookel2, line 249: Looking at the code from line 246 I think it would be more efficient to use an else here. E.g.
```suggestion
        if (collection.publishDate) {
            return date.format(collection.publishDate, "ddd, dd/mm/yyyy h:MMTT");
        } else {
            // This check should be a switch with manual but this
            // approach preserves the existing logic better in this
            // instance.
            if (collection.type === "automated") {
                return "[automated collection]";
            }

            return "[manual collection]";
        }
```

cookel2 (commented):
just a suggestion

src/app/views/collections/mapper/collectionMapper.js
lindenmckenzie: yep that's fair.

cookel2 (approved):
That works nicely 👍

PR #1241, opened 2026-06-15
Upgrade node to 24

dhee-tree:
### What

[Upgrade node to v24
](https://officefornationalstatistics.atlassian.net/browse/DIS-3998)

### How to review

- Tests pass
- Remove `node_modules` reinstall packages
```bash
cd src && rm -rf node_modules && npm install --legacy-peer-deps 
```
- Run service
- Create and publish a collection

### Who can review

!Me


ci/audit.yml
```diff
@@ -5,7 +5,7 @@ image_resource:
   type: docker-image
   source:
     repository: onsdigital/dp-concourse-tools-vulncheck-node
-    tag: latest
+    tag: 1.26.4-vulncheck-0.5.0-node-24
```
jessjenkins: We normally use `latest` for audit so as to make sure we are always using the latest version of vulncheck. Now this does present the quesiton of how we specify node 24 rather than 20 without breaking anything that needs to stay on node 20 for now. I wonder if we should just tag something like `latest-node-24` or similar?

dhee-tree: `latest-node-24` looks good to me, I will tag v24 with this and update the audit

dhee-tree: updated to use `latest-node-24`

jessjenkins (approved):

